PR #1, opened 2017-07-31
Enable access control

hasegeli:


serveradmin/dataset/commit.py
```diff
@@ -24,45 +24,60 @@ def __init__(self, commit, user):
         self.user = user
 
         if 'deleted' in commit:
-            self.deleted_servers = commit['deleted']
+            self.deleted = commit['deleted']
```
hasegeli: I am renaming this particular variable, though we would be able to rename the "deleted" on the client-server protocol.

serveradmin/dataset/commit.py
```diff
@@ -24,45 +24,60 @@ def __init__(self, commit, user):
...
         # If non-empty, the commit will go through the backend, but an error
         # will be shown on the client.
         self.warnings = []
 
+    def access_control(self, action, servers):
+        if not self.user or self.user.is_superuser:
+            return
+
+        access_control_groups = self.user.access_control_groups.all()
+        for server in servers.values():
```
hasegeli: I never learned docstring syntax.  I don't think it is a good investment to learn it now.  Python 3.5 has optional datatype safety [1].  We still have 3.4 on production, though.  We can wait a little longer and start using it.

[1] https://docs.python.org/3/library/typing.html

serveradmin/dataset/commit.py
```diff
@@ -155,39 +170,39 @@ def _delete_attributes(self):
...
-                del self.changed_servers[server_id]
+        for server_id in self.deleted:
+            if server_id in self.changes:
+                del self.changes[server_id]
 
     def _update_servers(self):
-        changed_servers = set()
-        for server_id, changes in self.changed_servers.items():
+        changes = set()
+        for server_id, changes in self.changes.items():
```
hasegeli: Nice catch!  I wonder how this was working.

serveradmin/dataset/commit.py
```diff
@@ -354,26 +375,18 @@ def commit_changes(
         )
 
 
-def _log_changes(deleted_servers, changed_servers, app, user):
-    # Import here to break cyclic import
```
hasegeli: Wasn't it relevant anymore.

serveradmin/dataset/commit.py
```diff
@@ -354,26 +375,18 @@ def commit_changes(
...
-    from serveradmin.dataset import query, filters
-
-    if deleted_servers:
-        old_servers = list(query(object_id=filters.Any(*deleted_servers)))
-    else:
-        old_servers = []
-
-    if not (changed_servers or old_servers):
+def _log_changes(deleted_servers, changes, app, user):
```
hasegeli: I am renaming them to `servers_to_change` and `servers_to_delete` as you suggested above.

serveradmin/dataset/create.py
```diff
@@ -70,6 +74,21 @@ def create_server(
         return server_id
 
 
+def access_control(user, server):
+    if not user or user.is_superuser:
```
hasegeli: Applications are also using this.  I will develop access control for them, later.

serveradmin/servershell/views.py
```diff
@@ -350,7 +345,6 @@ def get_values(request):
 
 
 @login_required     # NOQA: C901
-@permission_required('dataset.create_serverobject')
```
hasegeli: It is expected by the same function `_edit()`.

PR #2, opened 2017-07-31
Cleanup things around QuerySet()

hasegeli:


adminapi/dataset/filters.py
```diff
@@ -137,45 +148,35 @@ def __repr__(self):
...
-    'between': Between,
-    'not': Not,
-    'startswith': Startswith,
-    'overlap': Overlap,
-    'insidenetwork': InsideNetwork,
-    'privateip': PrivateIP,
-    'publicip': PublicIP,
-    'optional': Optional,
-    'empty': Empty,
+    d.lower(): globals()[d] for d in dir() if not d.startswith('_')
```
TheFRedFox: Should we mention that somewhere, that all classes are collected automatically? Could be dangerous in future, if one does not know about that.

adminapi/utils/parse.py
```diff
@@ -2,7 +2,7 @@ class ParseQueryError(Exception):
     pass
 
 
-def parse_function_string(args, strict=True):
+def parse_function_string(args, strict=True):   # NOQA: C901
```
TheFRedFox: 3 spaces to the comment. 2 are enough. 😉

serveradmin/dataset/filters.py
```diff
@@ -515,13 +480,23 @@ def from_obj(cls, obj):
...
+            )
+
         return _condition_sql(attribute, "{{0}} <<= ANY('{{{{{0}}}}}')".format(
             ','.join(str(n) for n in self.networks)
         ), servertypes)
 
 
 class InsideOnlyNetwork(InsideNetwork):
     def as_sql_expr(self, attribute, servertypes):
+        if attribute.type != 'inet':
```
TheFRedFox: Mhm, somehow it does not feel elegant, if you need to duplicate code now with the changes.

serveradmin/resources/views.py
```diff
@@ -15,7 +15,7 @@
 from serveradmin.serverdb.models import Project
 
 
-@login_required
+@login_required     # NOQA: C901
```
TheFRedFox: Why so many spaces until the comment?

TheFRedFox:
Just found some minor things. No code breaking things found.

serveradmin/resources/views.py
hasegeli: I align it to the next 4th.

adminapi/cmdline/query.py
```diff
@@ -25,9 +25,11 @@ def main():
 
     attrs = options.attrs if options.attrs else ('hostname', )
```
hasegeli: Then it wouldn't be a `tuple`.  A `list` is more appropriate in here anyway, so I am changing it.

adminapi/dataset/filters.py
hasegeli: I am adding a note about it.

adminapi/utils/parse.py
hasegeli: I align them to the next 4th column.

serveradmin/dataset/filters.py
hasegeli: I will clean it up again with the next one.  One of the purposes of this commit is to let the server-side filters go way completely.  Then we would have them all in a single place.

adminapi/cmdline/query.py
TheFRedFox: `('hostname',)` would not be a tuple anymore? I'm talking about just the space, not the comma here. ;)

serveradmin/dataset/filters.py
TheFRedFox: Fair enough then.

PR #4, opened 2017-09-12
Add automatic menu template loading

kofrezo:
This commit contains some improvements to ease getting started

- Add migrations because otherwise python manage.py migrate fails
- Ignore venv folder for virtual environement
- Try to automatically load menu.html templates and cache them


hasegeli:
Merged by hand

PR #5, opened 2017-09-14
Improve filters

hasegeli:


adminapi/base.py
```diff
@@ -0,0 +1,450 @@
...
+        self._server_object._save_old_value(self._attribute_id)
+        super(MultiAttr, self).intersection_update(new)
+
+    def difference_update(self, *others):
+        if not others:
+            return
+        new = others[0]
+        for other in others[1:]:
+            new |= other
+        if not self & new:
```
TheFRedFox: Maybe store the result first
``` python
new &= self
```
or the like, as I misread it first like
``` python
if not self and new:
    ...
```

btw. wouldn't it be the same here?

adminapi/base.py
```diff
@@ -0,0 +1,450 @@
...
+        new = other - self
+        if new:
+            MultiAttr._validate(new, self._attribute_id, self.datatype)
+        self._server_object._save_old_value(self._attribute_id)
+        super(MultiAttr, self).symmetric_difference_update(other)
+
+
+def _validate_value(attribute_id, value, datatype=None):
+    """It accepts an optional datatype to validate the values.  The values
+    are not necessarily be an instance of this datatype.  They will be checked
```
TheFRedFox: "... are not necessarily -be- an instance ..."

adminapi/base.py
```diff
@@ -0,0 +1,450 @@
...
+        self._order_by = attribute_ids
+        return self
+
+    def get_results(self):
+        if self._results is None:
+            self._fetch_results()
+        return self._results
+
+    def _fetch_results(self):
+        raise NotImplementedError()
```
TheFRedFox: Going through commit by commit right now. Will this be implemented later?

adminapi/base.py
```diff
@@ -0,0 +1,450 @@
...
+            obj.delete()
+        return self
+
+    def update(self, **attrs):
+        for obj in self:
+            if not obj.is_deleted():
+                obj.update(attrs)
+        return self
+
+    def iterattrs(self, attr='hostname'):
```
TheFRedFox: `iterattrs` would assume you can provide several attributes. Maybe `iterattr` would be better?

adminapi/base.py
```diff
@@ -0,0 +1,450 @@
...
+        value = self.get(key, default)
+        del self[key]
+        return value
+
+    def popitem(self):
+        for key in self.keys():
+            break
+        else:
+            raise KeyError()
+        return self.pop(key)
```
TheFRedFox: `return` can also be instead of the `break`.
Still hard to read for me, however, I guess it is the best way to do so.

adminapi/base.py
```diff
@@ -0,0 +1,450 @@
...
+            if not datatype and self.old_values.get(key) is not None:
+                datatype = type(self.old_values[key])
+            _validate_value(key, value, datatype)
+
+    def __delitem__(self, key):
+        self[key] = None
+
+    def clear(self):
+        for key in self:
+            del self[key]
```
TheFRedFox: Is it really meant to set every key to `None` or rather to clear also all the keys?

adminapi/base.py
```diff
@@ -0,0 +1,450 @@
...
+    def __setitem__(self, key, value):
+        if self._deleted:
+            raise DatasetError('Cannot set attributes to deleted server')
+        self.validate(key, value)
+        if isinstance(self[key], MultiAttr):
+            # Multi attributes are guaranteed to exist as MultiAttr, so
+            # we can always get the previous datatype from the existing
+            # value.
+            value = MultiAttr(value, self, key, self[key].datatype)
+        if key not in self.old_values or self.old_values[key] != value:
```
TheFRedFox: I don't think we need that if clause here. Half of it is already in the `_save_old_value` and the other half does not really make sense for me. Just try to save it.

adminapi/base.py
```diff
@@ -0,0 +1,450 @@
...
+        changes = {}
+        if self.is_dirty():
+            changes[self.object_id] = self._serialize_changes()
+        return {
+            'deleted': [self.object_id] if self.is_deleted() else [],
+            'changes': changes,
+        }
+
+    def _save_old_value(self, key):
+        if key not in self.old_values:
```
TheFRedFox: Consider to use `return-check` to reduce indentions.

adminapi/base.py
```diff
@@ -27,16 +27,9 @@ def __bool__(self):
...
-        kwargs = ', '.join(
-            '{0}={1!r}'.format(k, v) for k, v in self._filters.items()
-        )
-        query_repr = 'query({0})'.format(kwargs)
-
-        if self._restrict:
-            query_repr += '.restrict({0})'.format(', '.join(self._restrict))
-
-        return query_repr
+        return 'Query({{{}}})'.format(', '.join(
```
TheFRedFox: `{{{}}}` seriously? 3 nested format blocks? I even didn't try to understand that. ^^ What does it do? Why 3 nested blocks?

serveradmin/resources/views.py
```diff
@@ -7,7 +7,7 @@
 from django.views.decorators.csrf import ensure_csrf_cookie
 from django.conf import settings
 
-from adminapi.utils.parse import ParseQueryError, parse_query
+from adminapi.parse import ParseQueryError, parse_query
```
TheFRedFox: I guess you did rename `ParseQueryError` to `QueryError` and moved it to `adminapi.base` in a former commit, didn't you?

serveradmin/servershell/views.py
```diff
@@ -16,8 +16,8 @@
 from django.db import DataError, IntegrityError
 from django.utils.html import mark_safe, escape as escape_html
 
-from adminapi.utils.json import json_encode_extra
-from adminapi.utils.parse import ParseQueryError, parse_query
+from adminapi.parse import ParseQueryError, parse_query
```
TheFRedFox: Here too `adminapi.parse.ParseQueryError` vs `adminapi.base.QueryError`.

serveradmin/resources/views.py
TheFRedFox: https://github.com/innogames/serveradmin/pull/5/commits/b68f727d1a749a140632c39609c36895ed70fff7

Seems so, these changes should be in that commit too then. Seems you forgot to refactor these imports here then.

adminapi/cmduser.py
```diff
@@ -1,7 +1,26 @@
 import os
+from os.path import isfile
 import pwd
 
 
+def get_auth_token():
+    user = get_user()
+    config_file = os.path.join(user.pw_dir, '.adminapirc')
+    if isfile(config_file):
```
TheFRedFox: In a next move, you should try to reduce indentions.

serveradmin/serverdb/models.py
```diff
@@ -46,17 +46,17 @@
...
-    'reverse_hostname',
-    'number',
-    'inet',
-    'macaddr',
-    'date',
-    'supernet',
-)
+attribute_types = {
+    'string': str,
+    'boolean': lambda x: x in ('1', 'True', 'true', 1, True),
```
TheFRedFox: Why not using `strtobool` here again?

serveradmin/serverdb/models.py
```diff
@@ -403,8 +419,8 @@ def get_compiled_regexp(self):
     def get_default_value(self):
         if self.default_value:
             if self.attribute.multi:
-                return typecast(self.attribute, {self.default_value})
-            return typecast(self.attribute, self.default_value)
+                return self.attribute.typecast({self.default_value})
+            return self.attribute.typecast(self.default_value)
```
TheFRedFox: How about:

``` python
if not self.default_value:
    return self.attribute.initializer()()

value = self.default_value
if self.attribute.multi:
    value = {value}

return self.attribute.typecase(value)
```

Bit longer, but I guess more readable. ;)

adminapi/filters.py
```diff
@@ -1,5 +1,10 @@
-# This module must only include the filter classes.  Everything defined
-# in here are going to collected as the filters.
+from re import compile as re_compile, error as re_error
```
TheFRedFox: Seems not necessary as `re_compile` is the same as `re.compile` or is it because of performance?

adminapi/parse.py
```diff
@@ -97,24 +97,20 @@ def parse_query(term, filter_classes=fc, hostname=None):  # NOQA: C901
...
-                remaining_part, filter_classes, hostname_part
-            )
+            query_args = parse_query(remaining_part, None, hostname_part)
         else:
             hostname_part = term
             query_args = {}
 
         if any(x in hostname_part for x in _trigger_re_chars):
-            regexp_class = filter_classes['regexp']
+            regexp_class = Regexp
```
TheFRedFox: I guess without the dict lookup it does not make sense anymore to explicitly store the Regexp class.

serveradmin/servershell/views.py
```diff
@@ -165,12 +160,12 @@ def get_results(request):
...
     try:
-        query_args = parse_query(term, filter_classes)
-        q = Query(query_args).restrict('hostname')
+        query_args = parse_query(term)
+        query = Query(query_args).restrict('hostname')
     except (QueryError, ValidationError) as error:
         return HttpResponse(str(error), status=400)
 
-    hostnames = u' '.join(server['hostname'] for server in q)
+    hostnames = u' '.join(server['hostname'] for server in query)
```
TheFRedFox: `unicode` string should not be necessary, right?

serveradmin/servershell/views.py
```diff
@@ -17,6 +17,7 @@
 from django.utils.html import mark_safe, escape as escape_html
 
 from adminapi.base import QueryError
+from adminapi.filters import Any, InsideOnlyNetwork, Startswith
```
TheFRedFox: `filter_classes` was unnused before?

serveradmin/serverdb/querybuilder.py
```diff
@@ -1,31 +1,314 @@
...
+        sql += ' WHERE ' + self.get_sql_condition(
+            Attribute.objects.get(pk='servertype'),
+            Any(*(s.pk for s in self.servertypes)),
+        )
+        for attribute, value in self.filters.items():
+            sql += ' AND ' + self.get_sql_condition(attribute, value)
+
+        return sql
+
+    def get_sql_condition(self, attribute, value):  # NOQA C901
```
TheFRedFox: That's too much new stuff for to review in whole now. Would have suggest to do all this new stuff in a separated commit.

adminapi/filters.py
```diff
@@ -245,11 +234,8 @@ class InsideOnlyNetwork(InsideNetwork):
 
 
 class Empty(BaseFilter):
-    def __repr__(self):
-        return 'Empty()'
-
-    def serialize(self):
-        return {'name': 'empty'}
+    def __init__(self):
+        self.value = None
```
TheFRedFox: Wanna use the super function?
``` python
def __init__(self):
    super().__init__(None)
```
to not completely override but rather inherit the functionality?

hasegeli: I think it is better to keep it simple.  Filters are not something that would evolve.

serveradmin/serverdb/querybuilder.py
hasegeli: Fair enough.  It is kinda separate.  I will simplify this function later.

adminapi/parse.py
hasegeli: Fixed.

serveradmin/serverdb/models.py
hasegeli: Changed.

adminapi/cmduser.py
hasegeli: We can.

serveradmin/servershell/views.py
hasegeli: Fixed.

serveradmin/resources/views.py
hasegeli: The other way around, so it is fine.

adminapi/base.py
hasegeli: I added a comment.

adminapi/filters.py
```diff
@@ -78,12 +97,17 @@ def __repr__(self):
...
 
+    @classmethod
+    def deserialize_value(cls, value):
+        if not isinstance(value, list) or len(value) != 2:
+            raise FilterValueError(
+                'Invalid value for {}()'.format(cls.__name__)
+            )
+        return cls(value[0], value[1])
+
+    # TODO Remove
```
TheFRedFox: For other developers it could be interesting to know, why this should be removed.

You know, like you didn't get in touch with this code anymore and in 5 years another developers look at it and don't know why it should be removed.

adminapi/filters.py
```diff
@@ -102,11 +126,20 @@ def __repr__(self):
...
+        if not isinstance(value, list):
+            raise FilterValueError(
+                'Invalid value for {}()'.format(cls.__name__)
+            )
+        return cls(*value)
 
     def matches(self, value):
         return value in self.values
 
+    # TODO Remove
```
TheFRedFox: Also here.

adminapi/filters.py
```diff
@@ -126,14 +159,22 @@ def __repr__(self):
...
+        if not isinstance(value, list):
+            raise FilterValueError(
+                'Invalid value for {}()'.format(cls.__name__)
+            )
+        return cls(*[cls.deserialize(v) for v in value])
 
     def matches(self, value):
         return self.func(f.matches(value) for f in self.filters)
 
+    # TODO Remove
```
TheFRedFox: Also here.

adminapi/filters.py
TheFRedFox: Also:
* 142
* 177
* 212
* 227
* 261
* 308
* 314
* 78
* 56

serveradmin/servershell/templates/servershell/index.html
```diff
@@ -107,64 +107,14 @@ <h3>Available filters</h3>
 <table>
     <tr>
         <th>Name</th>
-        <th style="width:10em;">Arguments</th>
         <th>Description</th>
     </tr>
+    {% for name, description in filters %}
```
TheFRedFox: Nice! +1

serveradmin/serverdb/models.py
```diff
@@ -48,7 +49,7 @@
 
 attribute_types = {
     'string': str,
-    'boolean': lambda x: x in ('1', 'True', 'true', 1, True),
+    'boolean': strtobool,
```
TheFRedFox: Im really not sure, but wasnt there another location with something like that, too? Can't see it anymore here though.

serveradmin/serverdb/querybuilder.py
```diff
@@ -114,22 +114,16 @@ def get_sql_condition(self, attribute, filt):  # NOQA C901
...
-
+            template = "{{}} {} '{}'".format(operator, filt.value)
             if isinstance(filt, InsideOnlyNetwork):
                 template += (
                     ' AND NOT EXISTS ('
                     '   SELECT 1 '
                     '   FROM server AS supernet '
                     '   WHERE {{}} << supernet.intern_ip AND '
-                    '       supernet.intern_ip << ANY({})'
+                    '       supernet.intern_ip << {}'
```
TheFRedFox: Why are we using string concatenation for SQL queries in first place? Bet there is some better solution for that, right? Should have in mind for later to replace those concatenations with some query builder to reduce possibility of SQL injections.

serveradmin/servershell/views.py
```diff
@@ -17,7 +17,7 @@
 from django.utils.html import mark_safe, escape as escape_html
 
 from adminapi.base import QueryError
-from adminapi.filters import Any, InsideOnlyNetwork, Startswith, filter_classes
+from adminapi.filters import Any, ContainedOnlyBy, Startswith, filter_classes
```
TheFRedFox: Guess you also checked for all other places which `InsideOnlyNetwork` was used.

adminapi/filters.py
```diff
@@ -233,12 +233,19 @@ def from_obj(cls, obj):
         raise FilterValueError('Invalid object for Startswith')
 
 
-class Overlap(BaseFilter):
+class Overlaps(BaseFilter):
```
TheFRedFox: No imports to change?

serveradmin/common/templatetags/serversearch.py
```diff
@@ -22,7 +22,11 @@ def serversearch_js(search_id):
     return {
         'attributes_json': json.dumps(attributes),
         'filters_json': json.dumps(
-            [f.__name__ for f in filter_classes if f != ExactMatch]
+            [
+                f.__name__
+                for f in filter_classes
+                if f != ExactMatch and not f.__doc__.startswith('Deprecated')
```
TheFRedFox: Not like the idea to use such dynamic thing as docstring to filter deprecated stuff. Wonder if we could provide a decorator for that or just directly introduce a property to mark it as deprecated.

adminapi/filters.py
hasegeli: This commit changes the serialization format of the filters.  Those `from_obj()` methods are for backwards compatibility.  I will create another merge request to get rid of them immediately after merging this one.

serveradmin/servershell/templates/servershell/index.html
hasegeli: Yeah, finally up-to-date documentation!

serveradmin/serverdb/querybuilder.py
```diff
@@ -102,29 +102,7 @@ def get_sql_condition(self, attribute, filt):  # NOQA C901
...
-                template += (
-                    ' AND NOT EXISTS ('
-                    '   SELECT 1 '
-                    '   FROM server AS supernet '
-                    '   WHERE {{}} << supernet.intern_ip AND '
-                    '       supernet.intern_ip << {}'
-                    ')'
-                    .format(filt.value)
-                )
+            template = self._containment_filter_template(attribute, filt)
```
TheFRedFox: I know its just a separation, however, theoretically it does not fit with the commit message. 😜

serveradmin/serverdb/models.py
hasegeli: There is one more on `querybuilder.py`.  I will deal with that one by not casting but erroring out.  We first need to wait for everybody to upgrade though.  I put a `TODO` there for the time being.

serveradmin/serverdb/querybuilder.py
hasegeli: Certainly.  We need to change this to at least use parameterized queries.  Good thing is that after this merge request custom SQL generation is restricted to a single Python file only, admittedly not a simple one.

serveradmin/servershell/views.py
hasegeli: It was me who invented it and not exposed to the client side.  I am sure nobody else knows about it.

adminapi/filters.py
hasegeli: Possible.  It was again invented by me, exposed to the client side, but not documented.  I find it unlikely for people to rely on it.

serveradmin/common/templatetags/serversearch.py
hasegeli: It is just temporary.  The next merge request I promise will get rid of this as well.  I am putting a `TODO` in here too.

serveradmin/serverdb/querybuilder.py
hasegeli: Why not?

serveradmin/servershell/views.py
hasegeli: No, we are are running Python 3 anyway.  I am removing it.

serveradmin/serverdb/querybuilder.py
```diff
@@ -124,11 +116,11 @@ def _logical_filter_sql_condition(self, attribute, filt):
         else:
             joiner = ' OR '
 
-        if not filt.filters:
+        if not filt.values:
```
TheFRedFox: Hm, this change I don't understand. Where does these changes come from?

serveradmin/serverdb/querybuilder.py
TheFRedFox: Because the commit message just says about a Renaming of a class and not a separation and therefore addition of a method. However, I don't care about that here. Was just mentioning it. 😉

adminapi/filters.py
```diff
@@ -84,8 +84,36 @@ def from_obj(cls, obj):
...
+
+    def matches(self, value):
+        return self.value > value
+
+
+class LessThanOrEquals(BaseFilter):
+    """Filter the attribute less than or equals to the value"""
+
+    def matches(self, value):
+        return self.value < value
```
TheFRedFox: `<=`

adminapi/filters.py
```diff
@@ -158,10 +165,7 @@ def __repr__(self):
         )
 
     def serialize(self):
-        return {type(self).__name__: [
-            v.serialize() if isinstance(v, BaseFilter) else v
```
TheFRedFox: I know it is not possible anymore, however, maybe it would be good to keep that check to have a more defensive implementation?

TheFRedFox: Also (old lines):
* 176
* 218
* 227

serveradmin/dataset/queryset.py
```diff
@@ -88,15 +88,9 @@ def _get_query_builder(self):
...
             if attribute.pk == 'servertype':
-                if isinstance(filt, str):
-                    servertypes = {
-                        s for s in servertypes if s.pk == filt
-                    }
-                else:
-                    servertypes = {
-                        s for s in servertypes if filt.matches(s.pk)
-                    }
+                servertypes = {s for s in servertypes if filt.matches(s.pk)}
```
TheFRedFox: That is also possible because of the commit before, right?

serveradmin/serverdb/querybuilder.py
```diff
@@ -58,13 +56,14 @@ def get_sql_condition(self, attribute, filt):
         template = ''
 
         if attribute.type == 'boolean':
-            if isinstance(filt, BaseFilter):
+            if type(filt) != BaseFilter:
```
TheFRedFox: Don't really get that change? Why first check for BaseFilter and now contra BaseFilter?

TheFRedFox: Ah, also because of the commit before? But then why checking it here explicitly and removing all other checks?

serveradmin/serverdb/querybuilder.py
```diff
@@ -58,13 +56,14 @@ def get_sql_condition(self, attribute, filt):
...
                 raise FilterValueError(
                     'Boolean attribute "{}" cannot be used with {}() filter'
                     .format(attribute, type(filt).__name__)
                 )
 
-            negate = not filt or filt == 'false'
+            # TODO Better return errors for mismatching datatypes than casting
```
TheFRedFox: :+1:

serveradmin/serverdb/querybuilder.py
```diff
@@ -189,13 +188,10 @@ def _value_to_sql(self, attribute, value):
             except Server.DoesNotExist as error:
                 raise FilterValueError(str(error))
 
+        # TODO Better return errors for mismatching datatypes than casting
```
TheFRedFox: :+1: ^^

adminapi/filters.py
hasegeli: Fixed.

serveradmin/dataset/queryset.py
hasegeli: Yes.  It is the same commit actually.

serveradmin/serverdb/querybuilder.py
hasegeli: Now every filter passed to this function are `isinstance(BaseFilter)`.  For booleans attributes the the must exactly be `BaseFilter` not a subclass of it.  `BaseFilter` works for exact matching.  Other complicated filters doesn't make sense on booleans.

hasegeli: I am also adding a comment.

adminapi/base.py
```diff
@@ -0,0 +1,450 @@
...
+        super(BaseServerObject, self).__init__(attributes)
+        self.object_id = object_id
+        self._deleted = False
+        self._queryset = queryset
+        self.old_values = {}
+
+    def __hash__(self):
+        """Make the objects hashable
+
+        Note that it will not work for objects which doesn't have object_id.
```
HerrSpace: * don't have an object_id.

adminapi/base.py
HerrSpace: I don't think this is used anywhere in serveradmin directly so it's hard to tell, but I don't see when clearing the keys themselves would be usefull. Can you?

hasegeli: It doesn't look very useful to me either, given that you would never be able to commit this `Server` back to to Serveradmin.  Some attributes are surely required.  This is client-side code, so we would never know how is relying on it.  We can try removing it, see if anybody would complain.  I am appending another commit for it.

adminapi/base.py
hasegeli: Yes.  The client side and the server side implements it differently.

adminapi/base.py
hasegeli: I removed that one together with `__del__`.

adminapi/base.py
hasegeli: We need to save the first version only if got set multiple times.

hasegeli: On a second look, yes, it indeed doesn't make sense.

adminapi/base.py
hasegeli: This might even be wrong.  I am too tired to lazy to analyse this again.  I don't think anybody is using them anyway.  We gain so little by checking if the operation would change the values.  I will make them all just save the value and call the implementation.  We can later check later if the old values and new values are the same.  This would even cover the case when you set an attribute to some value and then set it back to its existing value.

TheFRedFox: Agree, gain would be too less to investigate if further.

PR #8, opened 2017-12-06
Retry adminapi requests before failing hard

meikomeik:
This commit will add a retry mechanism to adminapi requests.
If a request fails it will wait and try again three times before failing hard.

### DIFF
--- a/adminapi/request.py
+++ b/adminapi/request.py
@@ -35,25 +35,26 @@ def send_request(endpoint, data, auth_token, timeout=None):
         auth_token = get_auth_token()
 
     data_json = json.dumps(data, default=json_encode_extra)
-    try_backup = False
 
-    try:
-        req = _build_request(endpoint, auth_token, data_json, try_backup)
-        return json.loads(urlopen(req, timeout=timeout).read().decode('utf8'))
-    except HTTPError as error:
-        if error.code in (500, 502):
-            try_backup = True
-        else:
-            raise
-    except URLError:
-        try_backup = True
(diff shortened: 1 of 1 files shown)


adminapi/request.py
```diff
@@ -35,22 +35,25 @@ def send_request(endpoint, data, auth_token, timeout=None):
...
-        req = _build_request(endpoint, auth_token, data_json, try_backup)
-        return json.loads(urlopen(req, timeout=timeout).read().decode('utf8'))
-    except HTTPError as error:
-        if error.code in (500, 502):
-            try_backup = True
-        else:
-            raise
-    except URLError:
-        try_backup = True
+    while retries > 0:
```
hasegeli: `for retry in reversed(range(3))`

adminapi/request.py
```diff
@@ -35,22 +35,25 @@ def send_request(endpoint, data, auth_token, timeout=None):
...
 
-    if try_backup:
-        req = _build_request(endpoint, auth_token, data_json, try_backup)
-        return json.loads(urlopen(req, timeout=timeout).read().decode('utf8'))
+        # In case of an api error, sleep 5 seconds and try again three times
+        time.sleep(5)
+        retries -= 1
+
+    req = _build_request(endpoint, auth_token, data_json, True)
+    return json.loads(urlopen(req, timeout=timeout).read().decode('utf8'))
```
hasegeli: It would be nicer to eliminate this duplicate code now.  We can check do `if not retry: raise`.

PR #10, opened 2017-12-11
Remove deprecated filters

hasegeli:


adminapi/dataset/filters.py
```diff
@@ -1,16 +1,8 @@
 # XXX Deprecated
```
TheFRedFox: Will this whole module be removed at some later point?

hasegeli: Yes.  I incorporated it to this merge request as well.

PR #17, opened 2017-12-15
Clean up API further

hasegeli:


adminapi/request.py
```diff
@@ -42,7 +42,7 @@ def send_request(endpoint, data, auth_token, timeout=None):
             return json.loads(
                 urlopen(req, timeout=timeout).read().decode('utf8'))
         except HTTPError as error:
-            if error.code not in (500, 502):
+            if error.code < 500:
                 raise
             if retry == 0:
```
TheFRedFox: Defensive programming would go for `retry <= 0`, but not very important here.

adminapi/dataset/__init__.py
```diff
@@ -46,25 +40,23 @@ def commit(self, skip_validation=False, force_changes=False):
...
-            'filters': self._filters,
-            'restrict': self._restrict,
-            'order_by': self._order_by,
-        }
-        response = send_request(
-            QUERY_ENDPOINT, request_data, self.auth_token, self.timeout
-        )
-        return self._handle_response(response)
+    def get_results(self):
+        if self._results is None:
```
TheFRedFox: Is it possible that the response is handled as None as result? If so, it can be happen, that it will try to fetch the result everytime.

adminapi/request.py
```diff
@@ -17,30 +17,40 @@
...
         try:
-            req = _build_request(endpoint, auth_token, data_json)
+            req = _build_request(endpoint, Settings.auth_token, data_json)
             return json.loads(
-                urlopen(req, timeout=timeout).read().decode('utf8'))
+                urlopen(req, timeout=Settings.timeout).read().decode('utf8')
+            )
         except HTTPError as error:
             if error.code < 500:
                 raise
```
TheFRedFox: Here too, defensive programming would go for `retry <= 0` below.

TheFRedFox: Wonder if the `time.sleep` should also be configurable via `Settings`.

serveradmin/api/api.py
```diff
@@ -1,58 +0,0 @@
-import random
```
TheFRedFox: Seems like it would be just some debug functions? Would also be interesting to know why they were there and why you remove them now. 😉

adminapi/dataset/__init__.py
hasegeli: Of course, it is not possible.  They would be empty list after initialised.  That is why we check with `is None` in here.

adminapi/request.py
hasegeli: Done

adminapi/request.py
hasegeli: Done both

serveradmin/api/api.py
hasegeli: Me too.  I guess we will deploy and learn.

PR #18, opened 2017-12-27
 Support joins

hasegeli:


adminapi/datatype.py
```diff
@@ -4,6 +4,13 @@
 
 from netaddr import EUI
 
+# We use a set of regular expressions to cast to datatypes.  This module
+# is not aware of the attributes types of the server, neither it tries
```
TheFRedFox: `attribute types` or `attributes' types`?

TheFRedFox: Isn't it `neither .. nor`?

adminapi/api.py
```diff
@@ -37,28 +22,10 @@ def _api_function(*args, **kwargs):
 
             result = send_request(API_CALL_ENDPOINT, call)
 
-            if result['status'] == 'success':
-                return result['retval']
-
             if result['status'] == 'error':
+                raise ApiError(result['message'])
```
TheFRedFox: Maybe you can also provide information about the type of error in this `ApiError`. Passing `result['type']` to the `ApiError` would be worth considering.

adminapi/dataset/__init__.py
```diff
@@ -119,14 +119,13 @@ def commit(self, skip_validation=False, force_changes=False):
...
         result = send_request(COMMIT_ENDPOINT, commit)
 
-        if result['status'] == 'success':
-            self.num_dirty = 0
-            for obj in self:
-                obj._confirm_changes()
-            return True
-        elif result['status'] == 'error':
+        if result['status'] == 'error':
             _handle_exception(result)
```
TheFRedFox: Don't like the approach to use a method to raise exceptions. Would consider to let an exception to be returned an then raise it here.

TheFRedFox: Don't like it because it is hard to read. You can't see it directly that the code will be ended there for sure.

serveradmin/servershell/templates/servershell/index.html
```diff
@@ -7,40 +7,38 @@ <h2>Servershell</h2>
...
 </tr>
-<tr id="shell_understood_container">
-    <td>Understood: </td>
-    <td colspan="2" id="shell_understood">Nothing yet.</td>
-</tr>
 <tr id="shell_command_container">
-    <td>Command: </td>
+    <th>Command: </th>
     <td>
         <form method="post" action="javascript:return false;" id="shell_command_form">
```
TheFRedFox: What is that form good for here?

adminapi/dataset/__init__.py
```diff
@@ -35,6 +48,7 @@ def get_lookup(self, attr):
                 lookup[host[attr]] = host
         return lookup
 
+    # XXX: Deprecated
```
TheFRedFox: Isn't it ok to just keep these methods? Or should the Query be immutable?

adminapi/dataset/__init__.py
```diff
@@ -128,11 +135,12 @@ def commit(self, skip_validation=False, force_changes=False):
...
-            request_data = {
-                'filters': self._filters,
-                'restrict': self._restrict,
-                'order_by': self._order_by,
-            }
+            request_data = {'filters': self._filters}
+            if self._restrict is not None:
+                request_data['restrict'] = self._restrict
+            if self._order_by is not None:
+                request_data['order_by'] = self._order_by
```
TheFRedFox: Is it ok to change the structure of the `request_data` variable?

Instead of
```
{
    'filters': self._filters,
    'restrict': None,
    'order_by': None
}
```
it will now just be:
```
{'filters': self._filters}
```

serveradmin/serverdb/query_filterer.py
```diff
@@ -24,7 +24,7 @@
 from serveradmin.serverdb.models import Server, ServerAttribute
 
 
-class QueryBuilder(object):
+class QueryFilterer(object):
```
TheFRedFox: Why?

serveradmin/serverdb/query_filterer.py
```diff
@@ -43,6 +43,7 @@ def build_sql(self):
         )
         for attribute, value in self.filters.items():
             sql += ' AND ' + self.get_sql_condition(attribute, value)
+        sql += ' ORDER BY server.hostname, server.intern_ip, server.server_id'
```
TheFRedFox: Is it for sure that `hostname`, `intern_ip` and `server_id` are always present?

adminapi/dataset/__init__.py
```diff
@@ -437,7 +420,32 @@ def create(
...
+
+        dict.__setitem__(server_obj, attribute_id, casted_value)
+
+    return server_obj
+
 
-    return qs.get()
+def _format_attribute_value(value):
+    if isinstance(value, dict):
+        return _format_server(value)
```
TheFRedFox: Possible endless recursion (not really common in that case, but theoretically possible.

PR #27, opened 2018-03-20
Simplify adminapirc detection

eviscares:
The previous code misbehaves when you want to use roots token after
logging in with a personal ssh user.


hasegeli:
I got back on Serveradmin finally.  Thank you for cleaning this up.

PR #31, opened 2018-05-14
Show panel with development warning in debug mode

kofrezo:


serveradmin/settings.example.py
```diff
@@ -3,6 +3,7 @@
 import os
 
 DEBUG = True
+INTERNAL_IPS = ['127.0.0.1']
```
TuxPowered42: What about IPv6?

kofrezo: Added IPv6 localhost address.

PR #32, opened 2018-05-14
Make Servershell search bookmarkable

kofrezo:
Add a drop down menu to the servershell where one can select saved
searches.


hasegeli:
Why do we need this feature?  Can't you just get the URL and bookmark it?

hasegeli:
We should fix the current bookmark feature if it is not working, not add another one.

I fixed the "Graohite" typo.

PR #38, opened 2018-07-05
Stop caching lookup objects

hasegeli:


serveradmin/serverdb/query_materializer.py
```diff
@@ -113,24 +139,24 @@ def _add_attributes(self, servers_by_type):
...
             elif key == 'reverse':
                 reversed_attributes = {
                     a.reversed_attribute: a for a in attributes
                 }
                 for sa in ServerRelationAttribute.objects.filter(
                     value_id__in=self._server_attributes.keys(),
-                    _attribute__in=reversed_attributes.keys(),
-                ).select_related('server'):
+                    attribute__in=reversed_attributes.keys(),
+                ).select_related():
```
HerrSpace: Why do we not need the select_related('server') here anymore?

serveradmin/serverdb/query_materializer.py
```diff
@@ -261,7 +287,11 @@ def _get_join_results(self):
             for attribute, restrict in self._attributes.items():
                 if restrict is not None:
                     servers = self._get_servers_to_join(attribute)
-                    server_objs = type(self)(servers, restrict)
+                    server_objs = type(self)(
+                        servers,
+                        restrict,
+                        attribute_lookup=self._attribute_lookup,
+                    )
```
HerrSpace: Ahh, here attribute_lookup is used. Presumably just so we don't have to assemble it twice?

serveradmin/serverdb/sql_generator.py
```diff
@@ -211,7 +220,7 @@ def _containment_filter_template(attribute, filt):
     return template.format(_raw_sql_escape(value))
 
 
-def _condition_sql(servertypes, attribute, template):
+def _condition_sql(attribute, template, possible_servertype_ids=None):
     if attribute.special:
         field = attribute.special.field
         if field.startswith('_'):
```
HerrSpace: Is this still needed? This would match for queries like `_hostname=foobar`, right? When would that even happen?

serveradmin/serverdb/models.py
```diff
@@ -652,20 +478,16 @@ def save_value(self, value):
...
         on_delete=models.CASCADE,
         limit_choices_to=dict(type='boolean'),
     )
-    attribute = Attribute.foreign_key_lookup('_attribute_id')
 
     class Meta:
         app_label = 'serverdb'
         db_table = 'server_boolean_attribute'
-        unique_together = (('server', '_attribute'), )
-        index_together = (('_attribute', ), )
```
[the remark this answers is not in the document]
hasegeli: Oversight.  Fixing...

serveradmin/serverdb/query_executer.py
```diff
@@ -15,56 +16,105 @@
...
+    # Perhaps this is also enabling some optimization on the Postgres side.
     connection.cursor().execute(
         'SET TRANSACTION ISOLATION LEVEL REPEATABLE READ READ ONLY'
     )
-    servers = _get_servers(filters)
-    materializer = QueryMaterializer(servers, restrict, order_by)
 
-    return list(materializer)
+    # Query execution procedure is 2 steps: first filtering the object
+    # the materializing the requested attributes.  The joined attributes
```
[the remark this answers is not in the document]
hasegeli: Yes.

serveradmin/serverdb/query_materializer.py
hasegeli: `select_related()` without any arguments joins all direct relations to the query.  I had abused it to access the attribute objects.  This is wrong though.  Attributes should be passed there as objects already.  I am putting the `'server'` argument back there, and concentrating on reorganizing things more to preserve the Attribute objects.

serveradmin/serverdb/query_materializer.py
hasegeli: Yes, but I will move looking at attributes to the caller modules to make this nicer.

serveradmin/serverdb/sql_generator.py
hasegeli: Good catch!

adminapi/filters.py
```diff
@@ -77,6 +77,9 @@ def deserialize_value(cls, value):
     def matches(self, value):
         return value == self.value
 
+    def destiny(self):
+        return None
+
```
HerrSpace: These are very poetic.

PR #47, opened 2018-07-25
Move regexp from ServertypeAttribute to Attribute

HerrSpace:


serveradmin/serverdb/query_committer.py
```diff
@@ -435,18 +435,18 @@ def _validate_regexp(changes, servers, servertype_attributes):
         server = servers[object_id]
         for attribute_id, change in attribute_changes.items():
             sa = servertype_attributes[server['servertype']].get(attribute_id)
-            if not sa or not sa.regexp:
+            if not sa or not sa.attribute or not sa.attribute.regexp:
```
hasegeli: `sa.attribute` must be there.  We shouldn't check.

serveradmin/serverdb/query_committer.py
```diff
@@ -435,18 +435,18 @@ def _validate_regexp(changes, servers, servertype_attributes):
...
                 continue
 
             action = change['action']
             if action == 'update' or action == 'new':
                 if change['new'] is None:
                     continue
-                if not sa.regexp_match(change['new']):
+                if not sa.attribute.regexp_match(change['new']):
```
hasegeli: Accessing `sa.attribute` would cause a database query.  We should use the objects we already looked up.

serveradmin/serverdb/models.py
```diff
@@ -151,6 +166,29 @@ def from_str(self, value):
         except ValueError as error:
             raise ValidationError(str(error))
 
+    def get_compiled_regexp(self):
```
hasegeli: Shouldn't this be `_get_compiled_regexp`, or maybe `@property _compiled_regexp`?

HerrSpace:
I think this is ready to be reviewed again.

serveradmin/serverdb/query_committer.py
```diff
@@ -434,24 +434,26 @@ def _validate_readonly(attribute_lookup, changes, servers):
...
-def _validate_regexp(changes, servers, servertype_attributes):
+def _validate_regexp(changes, servers, attribute_lookup):
     for attribute_changes in changes:
         object_id = attribute_changes['object_id']
-        server = servers[object_id]
         for attribute_id, change in attribute_changes.items():
-            sa = servertype_attributes[server['servertype']].get(attribute_id)
-            if not sa or not sa.regexp:
+            if attribute_id in Attribute.specials:
+                continue
```
hasegeli: Shouldn't we validate special attributes as well?

HerrSpace: attribute_lookup does not contain special attributes. As far as I understand special attributes are regexp-validated by the django ORM via the 'validators' defined in the serverdb models.py. I'm not sure if we maybe bypass these validators here though.. 
Think the `if not sa or not sa.regexp` had the same effect before.

PR #51, opened 2018-08-02
Rework object_id handling

HerrSpace:
This fixes a bug where queries asking for a multi attribute, without asking for the members object_id caused an `TypeError: __hash__ method should return an integer` Error. Like:

```ruby
restrict = [{'acl_members': ['hostname', 'intern_ip']}]
```

### DIFF
--- a/adminapi/dataset/__init__.py
+++ b/adminapi/dataset/__init__.py
@@ -72,10 +72,17 @@ class BaseQuery(object):
             if restrict is None:
                 return None
 
+            if not isinstance(restrict, list):
+                raise TypeError('Restrict must be a list')
+
+            if 'object_id' not in restrict:
+                restrict.append('object_id')
+
             return [
-                {k: _ensure_object_id(v) for k, v in i.items()} if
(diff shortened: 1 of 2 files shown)


adminapi/dataset/__init__.py
```diff
@@ -269,7 +276,7 @@ def __init__(self, attributes=[]):
             if isinstance(value, (tuple, list, set, frozenset)):
                 attributes[attribute_id] = MultiAttr(value, self, attribute_id)
         super(DatasetObject, self).__init__(attributes)
-        self.object_id = self.get('object_id')
+        self.object_id = object_id
```
hasegeli: Shouldn't this be `_object_id`?  The external code should access `obj['object_id']`.

HerrSpace: obj['object_id'] is only there if 'object_id' was in restrict, while this object_id will always be there if the object is not new or deleted. We access that attribute in the graphite cache as well as in the servershell which makes it not really internal.

PR #55, opened 2018-10-12
Update Adminapi documentation

kofrezo:


docs/python-api.rst
```diff
@@ -80,28 +83,28 @@ committed.  To commit them you must call ``commit()`` on the query.
...
-for the existence of the ``game_function`` attribute::
+Another example will print all attributes of VM objects and check for the
+existence of the ``function`` attribute::
 
-    techerror = Query({'hostname': 'techerror.support.ig.local'}).get()
-    for attr, value in techerror.items(): # Iterate like a dict!
-         print "{0}={1}".format(key, value)
+    vm = Query().new_object('vm')
+    for attr in vm.keys():
+         print(attr)
```
HerrSpace: I think the key, value iteration part is pretty cool. We should keep that. Maybe you can combine that with your .new_object() example?

PR #56, opened 2018-10-22
Serveradmin packaging preparation

HerrSpace:
In preparation of deploying serveradmin as a package.


Pipfile
```diff
@@ -0,0 +1,49 @@
...
+"psycopg2-binary" = "*"
+# Django support for network address fields on PostgreSQL
+django-netfields = "*"
+# Required to generate sprites from Graphite
+Pillow = "*"
+# Used to list loaded api addons at /api/functions
+# TODO: Drop this dependency
+django-markup-deprecated = "*"
+# TODO: What is this for?
+docutils = "*"
```
hasegeli: Those 2 are probably not needed anymore.  The docs used to be rendered by Serveradmin, but I switched them to a link to GitHub.

serveradmin/__init__.py
```diff
@@ -3,3 +3,4 @@
 Copyright (c) 2018 InnoGames GmbH
 """
 
+VERSION = (1, 3, 1)
```
hasegeli: I think Serveradmin and Adminapi should share the version.

serveradmin/jinja2.py
```diff
@@ -1,17 +0,0 @@
...
-from django.core.urlresolvers import reverse
-
-
-def environment(**options):
-    env = Environment(**options)
-    env.globals.update({
-        'static': staticfiles_storage.url,
-        'url': reverse,
-    })
-    return env
```
hasegeli: `jinja2` was/is used by `ignetworking`.

Pipfile
HerrSpace: django-markup-deprecated is still needed at least. There is this page showing which api functions were loaded from external apps. That makes use of it. I'm not sure about docutils but I could check if that page still works without docutils. Thanks for the hint.

serveradmin/__init__.py
HerrSpace: They are the same as we can only build one package via the setup.py. This makes me question whether these two things should really live in the same repository. It doesn't seem worth the hassle to separate them though, that would be a PITA.

Moving this to the __init__.py of a module we will include in the package allows us to get this information during runtime when serveradmin was installed from a package. The setup.py is not available then while this would be in the python path.

serveradmin/jinja2.py
HerrSpace: This jinja2.py was copied straight from the django docs. It makes some more django env variables available in the templates like static and url. LB pushing doesn't use those though, I diffed the the config with and without this patch and there is no difference.

serveradmin/__init__.py
TheFRedFox: 👍  for separating the packages, but rather for somewhere in the future. ^^

serveradmin/api/urls.py
```diff
@@ -6,6 +6,7 @@
 from django.conf.urls import url
 
 from serveradmin.api.views import (
+    health_check,
```
TheFRedFox: Is Health Check really needed for this PR?

serveradmin/settings.py
```diff
@@ -121,117 +126,72 @@
...
     'graphOnly=true'
 )
-# User will be redirected to detailed system overview dashboard
-GRAFANA_DASHBOARD = (
-    'https://graphite.innogames.de/grafana'
-    '/dashboard/db/system-overview'
-)
+
+# Using exec certainly isn't an awesome solution but it's the best we've got.
+# The problem boils down to django configs beeing python files but python only
```
TheFRedFox: `beeing` => `being`

serveradmin/__init__.py
TheFRedFox: And 👍 for Gyros PITA!

serveradmin/api/urls.py
HerrSpace: Yes, as the new setup contains loadbalanced webservers which we didn't have before.

setup.py
```diff
@@ -1,26 +1,52 @@
...
-    entry_points={
-        'console_scripts': [
-            'adminapi=adminapi.cli:main',
-        ],
-    },
-)
+
+if __name__ == '__main__':
+    setup(
+        version='.'.join(str(d) for d in SERVERADMIN_VERSION),
```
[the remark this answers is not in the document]
HerrSpace: I like yours more.

serveradmin/settings.py
```diff
@@ -121,117 +126,72 @@
...
+# importing code from modules in its path. One solution would be to generate a
```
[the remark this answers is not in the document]
HerrSpace: I'll keep mine here.

serveradmin/settings.py
```diff
@@ -121,117 +126,72 @@
...
+# symlink while installing serveradmin, but I don't think there is a way to
```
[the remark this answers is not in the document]
HerrSpace: I'll keep mine here.

serveradmin/settings.py
```diff
@@ -121,117 +126,72 @@
...
+# isn't feasable as the symlink would be removed when upgrading the package
+# until config management has run again.
+dir_path = os.path.dirname(os.path.realpath(__file__))
+for config_path in [
+    os.environ.get('SERVERADMIN_CONFIGURATION'),
+    dir_path + '/local_settings.py',
+    '/etc/serveradmin/settings.py',
+]:
+    try:
+        if not config_path:
```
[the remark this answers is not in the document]
HerrSpace: Fixed!

serveradmin/settings.py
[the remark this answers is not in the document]
HerrSpace: Never mind ^^

PR #57, opened 2019-01-10
WIP: Support changing servertypes

RonjaPonja:
This adds support for changing the servertype via the normal commit API. It performs some sanity checking and finally modifies the commit to include attribute removals if an attribute was available on the old servertype and set but is not available on the new servertype. This guarantees no zombie attributes on the database and a clean history.

### DIFF
--- a/serveradmin/serverdb/query_committer.py
+++ b/serveradmin/serverdb/query_committer.py
@@ -59,10 +59,6 @@ def commit_query(created=[], changed=[], deleted=[], app=None, user=None):
     if not user:
         user = app.owner
 
-    pre_commit.send_robust(
-        commit_query, created=created, changed=changed, deleted=deleted
-    )
-
     entities = []
     if not user.is_superuser:
         entities.append(
@@ -80,6 +76,11 @@ def commit_query(created=[], changed=[], deleted=[], app=None, user=None):
         for a
         in list(attribute_lookup.values()) + list(Attribute.specials.values())
     }
+
+    _resolve_servertype_changes(changed, joined_attributes)
+    pre_commit.send_robust(
+        commit_query, created=created, changed=changed, deleted=deleted
+    )
 
     with transaction.atomic():
         change_commit = ChangeCommit.objects.create(app=app, user=user)
@@ -116,7 +117,8 @@ def commit_query(created=[], changed=[], deleted=[], app=None, user=None):
 
 
 def _validate(attribute_lookup, changed, changed_objects):
-    servertype_attributes = _get_servertype_attributes(changed_objects)
+    servertypes = {s['servertype'] for s in changed_objects.values()}
+    servertype_attributes = _get_servertype_attributes(servertypes)
 
     # Attributes must be always validated
     violations_attribs = _validate_attributes(
@@ -359,6 +361,12 @@ def _log_changes(commit, changed, created_objects, deleted_objects):
 
 
 def _fetch_servers(object_ids):
+    """ Fetch servers with row lock.
+
+    Returns mapping from object_id to a row in the Server table.  Raise an
+    error if an object in question (no longer) exists.  Lock the row of all
+    objects in question.
+    """
     servers = {
         s.server_id: s
         for s
@@ -373,15 +381,22 @@ def _fetch_servers(object_ids):
 
 
 def _materialize(servers, joined_attributes):
+    """ Get mapping from object_id to DatasetObject
+
+    QueryMaterializer will get the current values of joined_attributes and
+    resolve joins recursively.
+    """
     return {
         o['object_id']: o
         for o in QueryMaterializer(list(servers.values()), joined_attributes)
     }
 
 
-def _get_servertype_attributes(servers):
+def _get_servertype_attributes(servertypes):
+    """ Get mapping from servertype names to dicts of attributes
+    """
     servertype_attributes = dict()
-    for servertype_id in {s['servertype'] for s in servers.values()}:
+    for servertype_id in set(servertypes):
         servertype_attributes[servertype_id] = dict()
         for sa in Servertype.objects.get(pk=servertype_id).attributes.all():
             servertype_attributes[servertype_id][sa.attribute_id] = sa
@@ -397,11 +412,6 @@ def _validate_attributes(changes, servers, servertype_attributes):
         attributes = servertype_attributes[server['servertype']]
 
         for attribute_id, change in attribute_changes.items():
-            # If servertype is attempted to be changed, we immediately
-            # error out.
-            if attribute_id == 'servertype':
-                raise CommitValidationFailed('Cannot change servertype', [])
-
             # We have no more checks for the special attributes.
             if attribute_id in Attribute.specials:
                 continue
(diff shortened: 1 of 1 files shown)


serveradmin/serverdb/query_committer.py
```diff
@@ -116,7 +117,8 @@ def commit_query(created=[], changed=[], deleted=[], app=None, user=None):
 
 
 def _validate(attribute_lookup, changed, changed_objects):
-    servertype_attributes = _get_servertype_attributes(changed_objects)
+    servertypes = {s['servertype'] for s in changed_objects.values()}
```
hasegeli: I would call this `servertype_ids`.

serveradmin/serverdb/query_committer.py
```diff
@@ -359,6 +361,12 @@ def _log_changes(commit, changed, created_objects, deleted_objects):
 
 
 def _fetch_servers(object_ids):
+    """ Fetch servers with row lock.
```
hasegeli: Other docstrings don't have a space before and dot after the title.

serveradmin/serverdb/query_committer.py
```diff
@@ -359,6 +361,12 @@ def _log_changes(commit, changed, created_objects, deleted_objects):
...
+
+    Returns mapping from object_id to a row in the Server table.  Raise an
```
hasegeli: `s/mapping/dict`
`s/row/instance`
`s/table/model`

serveradmin/serverdb/query_committer.py
```diff
@@ -616,6 +626,91 @@ def _validate_real_attributes(servertype, real_attributes):     # NOQA: C901
...
+            'new' not in change['servertype'] or
+            not change['servertype']['new']
+        ):
+            CommitError('Cannot delete servertype')
+
+    # Get attributes of the new and old servertypes to check what changes
+    involved_servertypes = chain(*[
+        [change['servertype']['old'], change['servertype']['new']]
+        for change in servertype_changes
+    ])
```
hasegeli: No need to `*[]`.

serveradmin/serverdb/query_committer.py
```diff
@@ -616,6 +626,91 @@ def _validate_real_attributes(servertype, real_attributes):     # NOQA: C901
...
+
+    # Filter for servertype changes
+    servertype_changes = [
+        change for change in changed
+        if 'servertype' in change.keys()
+    ]
+
+    # No servertype changes, we're done here
+    if not servertype_changes:
+        return changed
```
hasegeli: Nobody checks the return value of this function.  You can just `return` to exit early.

serveradmin/serverdb/query_committer.py
```diff
@@ -616,6 +626,91 @@ def _validate_real_attributes(servertype, real_attributes):     # NOQA: C901
...
+
+    # The fetching and materializing of servers here is duplicating queries
+    # done in the commit_query transaction. We need to do this to resolve all
+    # the real attribute changes, resulting from a servertype change, before
+    # sending the pre_commit signal. Alternatively we could send the
+    # pre_commit signal inside the commit_query transaction but sending the
+    # signal can take many seconds if a receiver misbehaves.
+    with transaction.atomic():
+        changed_servers = _fetch_servers(
+            set(c['object_id'] for c in servertype_changes)
```
hasegeli: You could use set comprehension syntax.

serveradmin/serverdb/query_committer.py
```diff
@@ -616,6 +626,91 @@ def _validate_real_attributes(servertype, real_attributes):     # NOQA: C901
...
+        )
+    except Servertype.DoesNotExist:
+        raise CommitError('Unknown servertype')
+
+    # The fetching and materializing of servers here is duplicating queries
+    # done in the commit_query transaction. We need to do this to resolve all
+    # the real attribute changes, resulting from a servertype change, before
+    # sending the pre_commit signal. Alternatively we could send the
+    # pre_commit signal inside the commit_query transaction but sending the
+    # signal can take many seconds if a receiver misbehaves.
```
hasegeli: I couldn't get this.

serveradmin/serverdb/query_committer.py
```diff
@@ -616,6 +626,91 @@ def _validate_real_attributes(servertype, real_attributes):     # NOQA: C901
...
+    with transaction.atomic():
```
hasegeli: It is not okay to start another transaction in the same request.  If things fail after committing this one, you would end up persisting part of the `commit`.  Whatever critical need to go the `.atomic()` block started after calling this function.

hasegeli: Everything that write to the database is "critical".  There are more of them later on this function.

serveradmin/serverdb/query_committer.py
```diff
@@ -616,6 +626,91 @@ def _validate_real_attributes(servertype, real_attributes):     # NOQA: C901
...
+    # TODO: Handle attributes beeing required on the new servertype that are
+    # not available on the old servertype.  Changing servertypes means the
+    # available attributes change.  Adminapi and servershell are not equipped
+    # for this.  Hence we can not set values for these attributes in the same
+    # commit for now.  While this is not super clean, the user will be forced
+    # to set a correct value once he opens an edit dialog for this object in
+    # servershell.  I think that's good enough for now as changing servertypes
+    # is an edgecase anyway.
+
+    return None
```
hasegeli: No need

serveradmin/serverdb/query_committer.py
```diff
@@ -616,6 +626,91 @@ def _validate_real_attributes(servertype, real_attributes):     # NOQA: C901
...
+                }
+
+    # TODO: Handle attributes beeing required on the new servertype that are
+    # not available on the old servertype.  Changing servertypes means the
+    # available attributes change.  Adminapi and servershell are not equipped
+    # for this.  Hence we can not set values for these attributes in the same
+    # commit for now.  While this is not super clean, the user will be forced
+    # to set a correct value once he opens an edit dialog for this object in
+    # servershell.  I think that's good enough for now as changing servertypes
+    # is an edgecase anyway.
```
hasegeli: I am not convince it would be good enough.  Especially not having the defaults on the changed objects may cause real trouble.

hasegeli (changes requested):

serveradmin/serverdb/query_committer.py
HerrSpace: So I loop over all servertype_attributes of the new servertype and if it has a default value and the object does not have that attribute set, I set the default. Would that be enough?

serveradmin/serverdb/query_committer.py
HerrSpace: Fair enough, though mypy recommends this.

serveradmin/serverdb/query_committer.py
HerrSpace: I believe this function doesn't write to the database at all, except for the _fetch_servers row lock, which is why I added this transaction in the first place. I want this function to only expand the changes and that put the changes thorugh the normal commit process. Am I missing something here?

serveradmin/serverdb/query_committer.py
HerrSpace: Check out commit_query(). There we have the real transaction committing changes to the database. You chose to put the signals before and after the transactions as the signals are blocking until all receivers are done. To expand what attributes I have to delete I first have to materialize the object though. I need to expand the changes first as otherwise the pre_commit and post_commit hook would have different values for 'changed'. That would suck.

HerrSpace: So before my change the behavior was:
- pre_commit hook
- transaction
  - materialize the objects
  - commit the objects
- post_commit hook

Now the behavior is:
- transaction
  - materialize objects with changed servertypes
  - expand servertype changes to real attribute changes
- pre_commit hook
- transaction
  - materialize all objects
  - commit the objects
- post_commit hook

serveradmin/serverdb/query_committer.py
HerrSpace: Actually this doesn't seem to get flattened without `*[]` 🤔
```Internal Server Error: /servershell/commit
  File "/Users/space/.local/share/virtualenvs/serveradmin-m3Kn4UAw/lib/python3.5/site-packages/django/core/handlers/exception.py", line 41, in inner
    response = get_response(request)
  File "/Users/space/.local/share/virtualenvs/serveradmin-m3Kn4UAw/lib/python3.5/site-packages/django/core/handlers/base.py", line 249, in _legacy_get_response
    response = self._get_response(request)
  File "/Users/space/.local/share/virtualenvs/serveradmin-m3Kn4UAw/lib/python3.5/site-packages/django/core/handlers/base.py", line 187, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "/Users/space/.local/share/virtualenvs/serveradmin-m3Kn4UAw/lib/python3.5/site-packages/django/core/handlers/base.py", line 185, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/Users/space/.local/share/virtualenvs/serveradmin-m3Kn4UAw/lib/python3.5/site-packages/django/contrib/auth/decorators.py", line 23, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "./serveradmin/servershell/views.py", line 305, in commit
    commit_query(changed=changed, deleted=deleted, user=user)
  File "./serveradmin/serverdb/query_committer.py", line 80, in commit_query
    _resolve_servertype_changes(changed, joined_attributes)
  File "./serveradmin/serverdb/query_committer.py", line 663, in _resolve_servertype_changes
    involved_servertype_ids
  File "./serveradmin/serverdb/query_committer.py", line 398, in _get_servertype_attributes
    for servertype_id in set(servertype_ids):
TypeError: unhashable type: 'list'```

serveradmin/serverdb/query_committer.py
HerrSpace: fixed

serveradmin/serverdb/query_committer.py
HerrSpace: fixed

serveradmin/serverdb/query_committer.py
HerrSpace: True.

serveradmin/serverdb/query_committer.py
HerrSpace: fixed

serveradmin/serverdb/query_committer.py
HerrSpace: fixed

serveradmin/serverdb/query_committer.py
hasegeli: > otherwise the pre_commit and post_commit hook would have different values for 'changed'. That would suck.

I think it is fair that they are different.  Complicating the flow in here may have more downsides.

serveradmin/serverdb/query_committer.py
hasegeli: Does it still after you change the previous `return`?

HerrSpace: Honestly I currently don't even have it installed. I just made it a habit as mypy by default recommends to explicitly return everywhere, which I adapted. If I had it installed I would have caught the other return though ;) I'll just remove it for now and revisit once I get around to putting type hints in here.

PR #54, opened 2018-10-08
First steps for Django2 support

HerrSpace:

PR #60, opened 2019-03-08
Fix migrations

HerrSpace:
This PR borrows from https://github.com/innogames/serveradmin/pull/54

Historically our production DB has been migrated manually, not via django migrations. Further we didn't generate new migrations after model changes, but changed the initial migrations. After all of this the migrations were in a bad state.

With this MR I'm cleaning up this situation to allow for another model change I'm currently planing to add a datetime attribute datatype.

After this is merged we should diff the production db schema against these migrations and:
- clean up stuff left over on the production DB and
- add missing stuff to the models like indexes etc


serveradmin/apps/models.py
```diff
@@ -13,7 +13,7 @@ class Application(models.Model):
     name = models.CharField(max_length=80, unique=True)
     app_id = models.CharField(max_length=64, unique=True, editable=False)
     auth_token = models.CharField(max_length=64, unique=True, editable=False)
-    owner = models.ForeignKey(User)
+    owner = models.ForeignKey(User, on_delete=models.CASCADE)
```
TheFRedFox: Is it really meant that Applications will be deleted too when an user is deleted?
I rather think the Applications should stay with an undefined owner, right?

serveradmin/serverdb/migrations/0005_attribute_clone.py
```diff
@@ -10,11 +10,6 @@ class Migration(migrations.Migration):
     ]
 
     operations = [
-        migrations.AddField(
```
TheFRedFox: Was that field somewhere added before now? Because I could not find it and this migration would then fail, because of the RunSQL afterwards.

serveradmin/graphite/models.py
```diff
@@ -49,9 +49,10 @@ class Collection(models.Model):
     overview = models.BooleanField(default=False, help_text="""
         Marks the collection to be shown on the overview page.  For
         the overview page, sprites will be generated and cached on
-        the server in advance to improve the loading time.  {0} will be
-        appended to generated URL's to get the images for overview.
-        """.format(settings.GRAPHITE_SPRITE_PARAMS))
+        the server in advance to improve the loading time.  A suffix
+        will be appended to the generated URLs to get the overview
+        images, as defined by the GRAPHITE_SPRITE_PARAMS setting.
```
TheFRedFox: I dunno but was just wondering if it is possible to somehow make a link out of the `GRAPHITE_SPRITE_PARAMS`, so IDEs could find the reference.

serveradmin/apps/models.py
HerrSpace: Applications here is just serveradmins name for Application Tokens. Further I think we never delete users anyway but deactivate them only. Finally this is just explicitly stating the default on_delete behavior, as is required by django2. Even though we don't support django2 yet, this is part of the work.

serveradmin/serverdb/migrations/0005_attribute_clone.py
HerrSpace: It just moved to the initial migration.

serveradmin/serverdb/migrations/0003_server_indexing.py
```diff
@@ -15,10 +21,10 @@ class Migration(migrations.Migration):
             '       intern_ip inet_ops WITH &&,'
             '       servertype_id WITH ='
             '   )'
+            # At InnoGames we have a servertype loadbalancer which is the only
+            # servertype allowed to have multiple objects sharing the same
+            # intern_ip.  We use this to forward different ports to different
+            # hosts as well as for multi datacenter virtual IPs.
             "   WHERE (servertype_id != 'loadbalancer')"
```
hasegeli: This shouldn't be hardcoded, but use `ip_addr_type` of the servertype.  It is normally not allowed on Postgres for indexes to refer to other tables, but it'd work when the logic is wrapped into a function marked as `IMMUTABLE`.  The downside is that the changes on this column would corrupt the index.  I think it is unlikely for anybody to ever change `ip_addr_type` on an existing servertype.

Also the constraints can be made much better using such a function.  And I would also mark `ip_addr_type` field as not writable on Django admin at least to document this assumption.

Here is the SQL (not tested):

```sql
CREATE FUNCTION get_ip_addr_type(text)
    LANGUAGE sql STRICT IMMUTABLE
    AS 'SELECT ip_addr_type FROM servertype WHERE servertype_id = $1';

ALTER TABLE server
    ADD CONSTRAINT server_inter_ip_check
    CHECK ((get_ip_addr_type(servertype_id) = 'null') = (intern_ip IS NULL)),
    ADD CONSTRAINT server_host_inter_ip_exclude
    EXCLUDE (
        intern_ip WITH =
    )
    WHERE (get_ip_addr_type(servertype_id) = 'host'),
    ADD CONSTRAINT server_host_loadbalancer_inter_ip_exclude
    EXCLUDE (
        intern_ip WITH =,
        servertype_id WITH !=
    )
    WHERE (get_ip_addr_type(servertype_id) IN ('host', 'loadbalancer')),
    ADD CONSTRAINT server_network_inter_ip_exclude
    EXCLUDE USING gist (
        intern_ip inet_ops WITH &&,
        servertype_id WITH =
    )
    WHERE (get_ip_addr_type(servertype_id) = 'network');
```

HerrSpace: I agree, though this doesn't quite work yet.

I believe the function must define a return type like so:
```sql
CREATE FUNCTION get_ip_addr_type(text) RETURNS text
    LANGUAGE sql STRICT IMMUTABLE
    AS 'SELECT ip_addr_type FROM servertype WHERE servertype_id = $1';
```

The first two constraints seem fine, but the last two do not. I believe it's related to the inet datatype in combination with gist?

```
*serveradmin=# ALTER TABLE server
serveradmin-#     ADD CONSTRAINT server_host_loadbalancer_inter_ip_exclude
serveradmin-#     EXCLUDE (
serveradmin(#         intern_ip WITH =,
serveradmin(#         servertype_id WITH !=
serveradmin(#     )
serveradmin-#     WHERE (get_ip_addr_type(servertype_id) IN ('host', 'loadbalancer'));
ERROR:  operator <>(text,text) is not a member of operator family "text_ops"
DETAIL:  The exclusion operator must be related to the index operator class for the constraint.
```

```
*serveradmin=# ALTER TABLE server
serveradmin-#     ADD CONSTRAINT server_network_inter_ip_exclude
serveradmin-#     EXCLUDE USING gist (
serveradmin(#         intern_ip inet_ops WITH &&,
serveradmin(#         servertype_id WITH =
serveradmin(#     )
serveradmin-#     WHERE (get_ip_addr_type(servertype_id) = 'network');
ERROR:  data type character varying has no default operator class for access method "gist"
HINT:  You must specify an operator class for the index or define a default operator class for the data type.
```

Since this MR is only about cleaning up what's already there I think I will merge this so I can fake-migrate the whole thing once.

I'm interested in this change though and would like to keep track of it and create a new migration for it.

hasegeli: `btree` is not supporting `<>` operator.  `btree_gist` extension should support.

HerrSpace: Ah, I see, thanks. So does this look good?

```sql
CREATE FUNCTION get_ip_addr_type(text) RETURNS text
    LANGUAGE sql STRICT IMMUTABLE
    AS 'SELECT ip_addr_type FROM servertype WHERE servertype_id = $1';

ALTER TABLE server
    ADD CONSTRAINT server_inter_ip_check
    CHECK ((get_ip_addr_type(servertype_id) = 'null') = (intern_ip IS NULL)),
    ADD CONSTRAINT server_host_inter_ip_exclude
    EXCLUDE (
        intern_ip WITH =
    )
    WHERE (get_ip_addr_type(servertype_id) = 'host'),
    ADD CONSTRAINT server_host_loadbalancer_inter_ip_exclude
    EXCLUDE USING gist (
        intern_ip WITH =,
        servertype_id WITH !=
    )
    WHERE (get_ip_addr_type(servertype_id) IN ('host', 'loadbalancer')),
    ADD CONSTRAINT server_network_inter_ip_exclude
    EXCLUDE USING gist (
        intern_ip inet_ops WITH &&,
        servertype_id WITH =
    )
    WHERE (get_ip_addr_type(servertype_id) = 'network');

ALTER TABLE server DROP CONSTRAINT server_inter_ip_exclude;
```

HerrSpace: Applied, seems to work fine.

PR #61, opened 2019-03-08
serverdb: Add datetime attribute datatype

HerrSpace:
We want to change hypervisor locking from a boolean to a datetime field. We could just use an interger with a unix time stamp but having a real datetime allows us for nicer representations and datatype checking the adminapi client side.

This PR adds a new datatype for attributes.


serveradmin/serverdb/models.py
```diff
@@ -600,6 +603,34 @@ class Meta:
...
+        app_label = 'serverdb'
+        db_table = 'server_datetime_attribute'
+        unique_together = [['server', 'attribute', 'value']]
+        index_together = [['attribute', 'value']]
+
+    def save(self, *args, **kwargs):
+        # Our transport datetime format doesn't include a timezone.  Django
+        # will therefore create a python datetime instance without a timezone
+        # even though we use UTC by convention.  If local timezone usage is
+        # enabled in djangos config, it will also complain about the missing
```
hasegeli: Django's

adminapi/request.py
```diff
@@ -126,6 +127,8 @@ def _try_request(request, retry=False):
 def json_encode_extra(obj):
     if isinstance(obj, BaseFilter):
         return obj.serialize()
+    if isinstance(obj, datetime):
+        return obj.strftime('%Y-%m-%d %H:%M:%S')
```
hasegeli: We should append at least `+00` to this to make this clear.

adminapi/datatype.py
```diff
@@ -118,9 +120,14 @@ def str_to_datatype(value):
...
-            # date constructor is special.
+            # date constructors need a decode format
             if datatype is date:
-                return date(*(int(v) for v in value.split('-', 2)))
+                return datetime.strptime(value, "%Y-%m-%d").date()
+            if datatype is datetime:
+                parsed_datetime = datetime.strptime(value, "%Y-%m-%d %H:%M:%S")
+                # Our datetimes are UTC by convention. More information at:
+                # serveradmin.serverdb.ServerDateTimeAttribute.save
+                return parsed_datetime.replace(tzinfo=timezone.utc)
```
hasegeli: It'd be better to respect timezone offset passed in by the user.

serveradmin/serverdb/models.py
```diff
@@ -600,6 +603,34 @@ class Meta:
...
+    value = models.DateTimeField()
+
+    class Meta:
+        app_label = 'serverdb'
+        db_table = 'server_datetime_attribute'
+        unique_together = [['server', 'attribute', 'value']]
+        index_together = [['attribute', 'value']]
+
+    def save(self, *args, **kwargs):
+        # Our transport datetime format doesn't include a timezone.  Django
```
hasegeli: I couldn't get why.  As far as I understand, Django gives timezone aware datetimes.

HerrSpace: It's more about what we give django. Our transport format will be displayed pretty much 1:1 in servershell so it should be humanly accessible first. I asked @TuxPowered42 if we want timezones and or millisecond precision there and he said no to both.

HerrSpace: Please check again :)

adminapi/request.py
```diff
@@ -128,7 +128,9 @@ def json_encode_extra(obj):
     if isinstance(obj, BaseFilter):
         return obj.serialize()
     if isinstance(obj, datetime):
-        return obj.strftime('%Y-%m-%d %H:%M:%S')
+        if obj.tzinfo is None:
+            obj = obj.replace(tzinfo=timezone.utc)
```
hasegeli: Check, this is convenient.

adminapi/request.py
```diff
@@ -128,7 +128,9 @@ def json_encode_extra(obj):
...
+        return obj.astimezone(timezone.utc).strftime('%Y-%m-%d %H:%M:%S%z')
```
hasegeli: Why do you need this return.  Just casting it to string later would be fine, no?

adminapi/request.py
hasegeli: Although this would break Python 2 compatibility :(

HerrSpace: Fixed py2 compatibility. I understand why they didn't want to mess with keeping timzones up-to-date but not shipping an UTC timezone pre python3.2 is just dumb.

adminapi/request.py
HerrSpace: No. Casting to string would introduce milliseconds in the time and a colon in the utc offset like +01:00 instead of +0100. Pythons buildin strptime parsing is pretty rigid and doesn't allow for optional fields. We couldn't parse the result on the receiving side:

```python
> datetime.strptime('2019-03-23 18:29:08.374537+00:00', '%Y-%m-%d %H:%M:%S%z')

ValueError: time data '2019-03-23 18:29:08.374537+00:00' does not match format '%Y-%m-%d %H:%M:%S%z'
```

adminapi/request.py
HerrSpace: Eh, I found more problems. I'll check this later..

HerrSpace: Soo.. Running the following code breaks in py2:
```python
dt = datetime.utcnow().replace(tzinfo=utc)
dstr = dt.strftime('%Y-%m-%d %H:%M:%S%z')
datetime.strptime(dstr, '%Y-%m-%d %H:%M:%S%z')
```

The strftime actually works but the strptime doesn't know %z 🙄

HerrSpace: Maybe this is the time to finally drop python2 support. Riding a dead horse.

PR #69, opened 2019-07-09
Fix broken ACL evaluation

HerrSpace:
### Improves ACL docs and error messages:
- Before we only informed the user that its ACLs don't allow a change.
- Now we return information about why each ACL didn't allow a change.

### Allows attribute blacklists instead of whitelists in ACLs:
- Before the list of attributes in an ACL was a whitelist
- Now it can be toggled between being a whitelist and a blacklist

This addresses two separate problems:
- When adding a new attribute to a servertype, it is missing from all ACL whitelists.  Some applications, like the GOP functional tests require deleting objects in their project.  Hence their ACL must allow all attributes.  Each time a new attribute is added to VM the ACLs need to be adjusted or deleting the VMs will no longer be allowed.  _That said each ACL should probably only allow a small set of attributes to be changed on a large set of hosts (backup) or a large set of attributes on a small set of hosts (functional tests).  Never both._
- Only real attributes can be selected in the ACLs attribute lists, special attributes (hostname, intern_ip...) can not, because they are not present in the attributes table.  In blacklist mode we now allow `(normal attributes + special attributes) - acl attributes`.  Of course you also can't blacklist special attributes now so it's kind of a hack.

### Fixes a vulnerability allowing the change of non-whitelisted attributes:
- Before we compared the new attribute value against, again, the new attribute value.  Obviously this check always returns equality and allowed the change.


serveradmin/serverdb/query_committer.py
```diff
@@ -292,43 +285,125 @@ def _upsert_attributes(attribute_lookup, changed, changed_servers):
...
+
+            # If all ACLs resulted in violations, none of them allowed the edit
+            # Build a verbose error message and abort the commit
+            if all(acl_violations.values()):
+                msg = (
+                    'Insufficient access rights to object "{}" for {} "{}": '
+                    .format(obj['hostname'], entity_class, entity_name)
                 )
+                for acl, violations in acl_violations.items():
+                    msg += ' '.join(violations)
```
kofrezo: Maybe add newlines here to increase readability of the error messages ?

HerrSpace: I actually tried this initially but the django message box thingy we use collapses newlines back into spaces anyway. Further we also log this message and multi line log messages could be hard to make sense of.

TuxPowered42: I had similar issues with errors from HWLB pushing and I went with `<pre>` tag around output. Maybe that would be an option?

PR #71, opened 2019-07-18
Add last login tracking to Applications

HerrSpace:


serveradmin/apps/migrations/0004_application_last_login.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('apps', '0003_public_key_length'),
+    ]
+
+    operations = [
+        migrations.AddField(
```
TheFRedFox: What is the reason we do not store every single login but "just" the last one?

serveradmin/apps/models.py
```diff
@@ -34,6 +34,7 @@ class Application(models.Model):
     owner = models.ForeignKey(User, on_delete=models.CASCADE)
     location = models.CharField(max_length=150)
     disabled = models.BooleanField(default=False)
+    last_login = models.DateTimeField(null=True, default=None, editable=False)
```
TheFRedFox: Help me here quick, what does `editable=False` do again?

serveradmin/apps/migrations/0004_application_last_login.py
HerrSpace: See: https://github.com/innogames/serveradmin/pull/71/files#diff-e62344281bdb7c2545658a551867dc62R142

We don't have sessions so each request is authenticated. Some apps like puppet send thousands of request per second. Each of them would cause an additional database commit and hence an additional flush to disk. That's pretty expensive. We only care about which apps we can disable here anyway. If you want that information you can look into the logs / send them to kibana.

serveradmin/apps/models.py
HerrSpace: The field becomes non-editable in the autogenerated admin ui.

TheFRedFox: Nice. Thx. ^^

serveradmin/apps/migrations/0004_application_last_login.py
TheFRedFox: Makes absolutely sense. Good to go.

PR #79, opened 2020-02-18
fix(serverdb): Allow two or more hyphens in hostname

kofrezo:
According to RFC952 and RFC1123 having two or more hyphens in a row for
labels as long as they don't start with them.


serveradmin/serverdb/models.py
```diff
@@ -49,7 +49,7 @@
 
 HOSTNAME_VALIDATORS = [
     RegexValidator(
-        r'\A(\*\.)?([a-z0-9]+[\.\-])*[a-z0-9]+\Z', 'Invalid hostname'
+        r'\A(\*\.)?([a-z0-9]+([\.]|[\-]*))*[a-z0-9]+\Z', 'Invalid hostname'
```
meikomeik: I think we can remove a few square brackets and we don't need to escape the hyphen.
```suggestion
        r'\A(\*\.)?([a-z0-9]+(\.|-+))*[a-z0-9]+\Z', 'Invalid hostname'
```

kofrezo: True, I will adjust the migration script.

TheFRedFox: Good suggestion, @meikomeik. Especially the `+` after the hyphen was important to avoid ReDOS attacks.

PR #80, opened 2020-03-09
Rewrite Servershell Javascript and use Boostrap HTML toolkit

kofrezo:
Make Serveradmin use the popular HTML toolkit Boostrap in its latest version 4.4 for the following reasons:
- Provide a modern and familiar look & feel
- Provide a framework that supports all screens & sizes (responsive grid system)
- Allow others to easily extend Serveradmin with less effort (well known, docs online available)
- Update javascript libraries to the latest version (e.g. jQuery)

The Servershell Javascript has been fully rewritten and split up to
- Reduce complexity by splitting the code into multiple parts
- Provide a rich documentation to easily understand why things are happening
- Provide a platform for future testing
- Use modern framework functions to avoid writing custom code

Additionally some pages related to the Servershell such as graphite graphs, history editing etc. have been fixed or extended capabilities.

List of noticeable changes:

- Move navigation bar from left side to top
- Add a spinner to the top right to show pending actions (running ajax calls)
- Add legend for attributes letter codes
- Automatically retry commit and search if backend request fails
- Collapsible menu for search settings and keyboard shortcuts legend
- Allow user to en/disable auto completion for search and command
- Allow user to en/disable auto select first element in auto completion for search and command
- Allow tab control in forward and reverse order (search, command, result)
- Allow alt + backspace to delete words (same as in shell)
- Tabbing from search to command input submits search (additionally to pressing enter)
- Provide auto completion for attribute values (e.g. game_market=d -> de, default)
- Save selection of shown attributes across searches and tabs
- Allow to choose fields to export (export command)
- Allow to filter for attribute (history command)
- Show current per page setting in result info
- Auto select first entry of result list if there is only one
- Auto select edited fields
- On-the-fly validation for inline editing values in the result table and edit page
- Add alerts for info, warning and error messages (e.g. commit failed, wrong usage of command)
- Open all actions (e.g. history, edit, new) in same or new tab instead of modals (except IP choosing)
- Allow complex filtering on changes page (from, to, hostname, object id, user or app)
- Make changes page more readable (relative timestamp, show hostname, group changes)
- Fix restore object option for deleted objects (create new object of copying old data)
- Allow to search for text in the history page (history command link to here)

List of things that got away or still have to be done:
- The hidden feature to use shift arrow keys to access the command history is currently missing
- Compress and minify javascript files for production
- The changes page is still slow due to many joins and fetches


serveradmin/servershell/helper/autocomplete.py
```diff
@@ -0,0 +1,98 @@
...
+    :param attribute_id: e.g. servertype
+    :param search_string: e.g. v
+    :param limit: limit result to n results
+    :return:
+    """
+
+    column_name = Attribute.specials[attribute_id].special.field
+
+    with connection.cursor() as cursor:
+        # This is safe because field_name comes from us
```
0xBADBAC0N: Even if these fields come from us, why do we not use a parameterized query and prepared statement? :)

serveradmin/servershell/static/js/servershell/command.js
```diff
@@ -0,0 +1,542 @@
...
+ * attribute or a normal single value attribute. It will detect if multi
+ * changes nullify each other (mostly) and remove the changes to commit
+ * no changes to backend.
+ *
+ * This method does NOT validate anything in any way. You must have checked
+ * if the object, attribute and values before hand.
+ *
+ * @param object_id e.g. 123456
+ * @param attribute_id  e.g. state or responsible_admin
+ * @param new_value e.g. maintenance or ['daniel.kroeger', 'frederik.bauer']
```
0xBADBAC0N: Remove PII, in special from public Github :)

serveradmin/servershell/views.py
```diff
@@ -1,90 +1,129 @@
 """Serveradmin - Servershell
 
-Copyright (c) 2019 InnoGames GmbH
+Copyright (c) 2020 InnoGames GmbH
```
0xBADBAC0N: To save some work once a year ;D
`date.today().year`

serveradmin/servershell/helper/autocomplete.py
kofrezo: Because the execute function blindly escapes all occurrences of %s with two single quotes to make sure it is 100% robust against SQL injections. This will break the SQL query as you the parameters in the select and where clause are column names and the limit must be a integer.
`
sql = "SELECT DISTINCT %s FROM server WHERE %s LIKE %s ORDER BY %s ASC LIMIT %s"
c.execute(sql, ['hostname', 'hostname', 'de100%', 'hostname', '20'])
django.db.utils.ProgrammingError: non-integer constant in ORDER BY
LINE 1: ...OM server WHERE 'hostname' LIKE 'de100%' ORDER BY 'hostname'...
`

serveradmin/servershell/views.py
kofrezo: Because it harms readability to have doc strings concatenated with " + date.today().year + " and makes it possibly broken for automatic doc generators.

kofrezo: I will close it for now. Feel free to get in touch with me. Thanks for the review btw!

PR #82, opened 2020-04-01
Minor Servershell + Resources bug fixes

kofrezo:
- Fix pagination not working on resources page
- Fix using pagination override deep link on resources page
- Fix per_page filter not applied on resources page
- Fix search term not the same on resources page and Servershell search
- Fix search term not saved in session
- Fix typing too fast omits characters

### DIFF
--- a/serveradmin/resources/templates/resources/index.html
+++ b/serveradmin/resources/templates/resources/index.html
@@ -40,7 +40,7 @@
                     <label for="per_page" class="col-sm-1 col-form-label">Per Page:</label>
                     <div class="col-md-4">
                         <input class="form-control form-control-sm" id="per_page" name="per_page" type="number" value="{{ per_page }}" />
-                        <input type="hidden" id="page" name="page" value="1" />
+                        <input type="hidden" id="page" name="page" value="{{ page }}" />
                     </div>
                 </div>
             </div>
--- a/serveradmin/resources/templates/resources/index.html
+++ b/serveradmin/resources/templates/resources/index.html
@@ -39,7 +39,7 @@
                 <div class="form-group row input-controls">
                     <label for="per_page" class="col-sm-1 col-form-label">Per Page:</label>
                     <div class="col-md-4">
-                        <input class="form-control form-control-sm" id="per_page" type="number" value="{{ per_page }}" />
+                        <input class="form-control form-control-sm" id="per_page" name="per_page" type="number" value="{{ per_page }}" />
                         <input type="hidden" id="page" name="page" value="1" />
                     </div>
                 </div>
--- a/serveradmin/resources/templates/resources/index.html
+++ b/serveradmin/resources/templates/resources/index.html
@@ -9,7 +9,7 @@
 {% endblock %}
 
 {% block content %}
-<form method="get" action="{% url 'resources_index' %}">
+<form id="resources-form" method="get" action="{% url 'resources_index' %}">
     <div id="filterCollapse" class="collapse">
         <div class="row">
             <div class="col-md-12 controls">
@@ -40,6 +40,7 @@
                     <label for="per_page" class="col-sm-1 col-form-label">Per Page:</label>
                     <div class="col-md-4">
                         <input class="form-control form-control-sm" id="per_page" type="number" value="{{ per_page }}" />
+                        <input type="hidden" id="page" name="page" value="1" />
                     </div>
                 </div>
             </div>
@@ -113,17 +114,15 @@
         <div class="pagination">
             <span class="step-links">
                 {% if hosts.has_previous %}
-                    <a href="?page=1">&laquo; first</a>
-                    <a href="?page={{ hosts.previous_page_number }}">previous</a>
+                    <a href="#" onclick="$('#page').val($(this).data('page')); $('#resources-form').submit();" data-page="1">&laquo; first</a>
+                    <a href="#" onclick="$('#page').val($(this).data('page')); $('#resources-form').submit();" data-page="{{ hosts.previous_page_number }}">previous</a>
                 {% endif %}
-
                 <span class="current">
                     Page {{ hosts.number }} of {{ hosts.paginator.num_pages }}.
                 </span>
-
                 {% if hosts.has_next %}
-                    <a href="?page={{ hosts.next_page_number }}">next</a>
-                    <a href="?page={{ hosts.paginator.num_pages }}">last &raquo;</a>
+                    <a href="#" onclick="$('#page').val($(this).data('page')); $('#resources-form').submit();" data-page="{{ hosts.next_page_number }}">next</a>
+                    <a href="#" onclick="$('#page').val($(this).data('page')); $('#resources-form').submit();" data-page="{{ hosts.paginator.num_pages }}">last &raquo;</a>
                 {% endif %}
             </span>
         </div>
--- a/serveradmin/resources/templates/resources/index.html
+++ b/serveradmin/resources/templates/resources/index.html
@@ -39,7 +39,7 @@
                 <div class="form-group row input-controls">
                     <label for="per_page" class="col-sm-1 col-form-label">Per Page:</label>
                     <div class="col-md-4">
-                        <input class="form-control form-control-sm" id="per_page" name="per_page" type="number" value="{{ per_page }}" />
+                        <input class="form-control form-control-sm" min="1" id="per_page" name="per_page" type="number" value="{{ per_page }}" />
                         <input type="hidden" id="page" name="page" value="{{ page }}" />
                     </div>
                 </div>
--- a/serveradmin/resources/views.py
+++ b/serveradmin/resources/views.py
@@ -30,8 +30,7 @@ def index(request):
     term = request.GET.get('term', request.session.get('term', ''))
     collections = list(Collection.objects.filter(overview=True))
 
-    # If a graph collection was specified, use it.  Otherwise use the first
-    # one.
+    # If a graph collection was specified, use it. Otherwise use the first one
     for collection in collections:
         if request.GET.get('current_collection'):
             if str(collection.id) != request.GET['current_collection']:
@@ -111,13 +110,15 @@ def index(request):
             hosts[server['hostname']] = dict(server)
 
     page = int(request.GET.get('page', 1))
-    per_page = int(request.GET.get('per_page', 8))
+    per_page = int(request.GET.get(
+        'per_page', request.session.get('resources_per_page', 8)))
+    request.session['resources_per_page'] = per_page
     hosts_pager = Paginator(list(hosts.values()), per_page)
 
     sprite_url = settings.MEDIA_URL + 'graph_sprite/' + collection.name
     template_info.update({
         'columns': columns,
-        'hosts': hosts_pager.page(1),
+        'hosts': hosts_pager.page(page),
         'page': page,
         'per_page': per_page,
         'matched_hostnames': matched_hostnames,
--- a/serveradmin/resources/views.py
+++ b/serveradmin/resources/views.py
@@ -29,7 +29,7 @@ def index(request):
     term = request.GET.get('term', request.session.get('term', ''))
     collections = list(Collection.objects.filter(overview=True))
 
-    # If a graph collection was specified, use it. Otherwise use the first one
+    # If a graph collection was specified, use it.  Otherwise use the first one
     for collection in collections:
         if request.GET.get('current_collection'):
             if str(collection.id) != request.GET['current_collection']:
--- a/serveradmin/resources/views.py
+++ b/serveradmin/resources/views.py
@@ -7,8 +7,8 @@ from collections import OrderedDict
 
 from django.conf import settings
 from django.contrib.auth.decorators import login_required
-from django.core.exceptions import ValidationError
-from django.core.paginator import Paginator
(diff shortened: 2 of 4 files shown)


serveradmin/resources/templates/resources/index.html
```diff
@@ -39,7 +39,8 @@
                 <div class="form-group row input-controls">
                     <label for="per_page" class="col-sm-1 col-form-label">Per Page:</label>
                     <div class="col-md-4">
-                        <input class="form-control form-control-sm" id="per_page" type="number" value="{{ per_page }}" />
+                        <input class="form-control form-control-sm" id="per_page" name="per_page" type="number" value="{{ per_page }}" />
+                        <input type="hidden" id="page" name="page" value="1" />
```
TheFRedFox, line 43:
```suggestion
                        <input type="hidden" id="page" name="page" value="{{ page }}" />
```
so, when accidentally submitting the form manually, it won't bring you back to the first page.

serveradmin/resources/views.py
```diff
@@ -30,8 +30,7 @@ def index(request):
     term = request.GET.get('term', request.session.get('term', ''))
     collections = list(Collection.objects.filter(overview=True))
 
-    # If a graph collection was specified, use it.  Otherwise use the first
-    # one.
+    # If a graph collection was specified, use it. Otherwise use the first one
```
TheFRedFox, line 33: Actually PEP8 states:
https://www.python.org/dev/peps/pep-0008/#comments
> You should use two spaces after a sentence-ending period in multi- sentence comments, except after the final sentence.

serveradmin/resources/views.py
```diff
@@ -111,13 +110,15 @@ def index(request):
             hosts[server['hostname']] = dict(server)
 
     page = int(request.GET.get('page', 1))
```
TheFRedFox, line 112: What will the paginator actually do if `page` is negative or 0?

serveradmin/resources/views.py
```diff
@@ -111,13 +110,15 @@ def index(request):
...
-    per_page = int(request.GET.get('per_page', 8))
+    per_page = int(request.GET.get(
```
TheFRedFox, line 113: Same here, what will happen if per_page is negative or 0?

serveradmin/servershell/static/js/servershell/autocomplete/search.js
```diff
@@ -16,8 +16,7 @@ $(document).ready(function () {
 
     let term_input = $('#term');
     term_input.autocomplete({
-        delay: 50, // Wait n ms before starting to auto complete to avoid needles requests to backend
-        minLength: 0,
```
TheFRedFox, line 20: what's the default minLength? Good that it's not `minLength: 0` anymore. ^^

serveradmin/servershell/views.py
```diff
@@ -162,6 +162,9 @@ def get_results(request):
             'message': str(error)
         }))
 
+    # Query successful term must be valid we can save it.
```
TheFRedFox, line 165: This sentence is not completely understandable for me. I guess you meant it like that:
```suggestion
    # Query successful term must be valid here, so we can save it safely now.
```

serveradmin/resources/views.py
kofrezo: I did the following now:
You can not enter a invalid page number using the UI. If the user has manually submitted a invalid page with the Query string he probably wants something harmful. We raise a SuspiciousOperation exception now which leads to a 400 - Bad Request

serveradmin/resources/views.py
kofrezo: I set a min value so that it is not possible. If the user does so by manipulating the URL nothing happens but the paginator show a negative value.

serveradmin/servershell/static/js/servershell/autocomplete/search.js
kofrezo: The default value is 1. I think 0 had a purpose for users knowing there is no auto completion. They just enter the field and immediately got some help but this is probably better.

serveradmin/resources/views.py
```diff
@@ -113,19 +113,24 @@ def index(request):
     per_page = int(request.GET.get(
         'per_page', request.session.get('resources_per_page', 8)))
     request.session['resources_per_page'] = per_page
-    hosts_pager = Paginator(list(hosts.values()), per_page)
+
+    try:
+        hosts_pager = Paginator(list(hosts.values()), per_page).page(page)
```
TheFRedFox, line 118: Hm, also possible. I was rather thinking of just setting `page` and `per_page` to a default as well when it is not an integer and negative.

serveradmin/resources/views.py
```diff
@@ -113,19 +113,24 @@ def index(request):
...
+    except (PageNotAnInteger, EmptyPage):
```
TheFRedFox, line 119: `EmptyPage` .. what about queries which actually results into no entries?

serveradmin/resources/views.py
TheFRedFox: A max limit might also make sense, to not make too big queries, if you can already define a max here.

serveradmin/resources/views.py
TheFRedFox: Never trust User Input. UI is nice but can always be bypassed. Changing html or just building the request on your own. So backend validation is the right thing to go. ;)

serveradmin/resources/views.py
kofrezo: > Hm, also possible. I was rather thinking of just setting `page` and `per_page` to a default as well when it is not an integer and negative.

I see your point but I think if a user somewhat entered invalid input here there are only 2 cases.

First one: 
There is a bug in the UI and we should fix it in a way the user can not.

Second one:
The user entered invalid data on purpose and we should make this clear to him by showing an error and notify a developer (by exception).

I don't like the idea of not telling the user that the input was wrong and instead behaving in a way one would not expect or would not match the input but the default.

> A max limit might also make sense, to not make too big queries, if you can already define a max here.

I don't have any it would be arbitrary we have seen 100 hosts working just fine.

serveradmin/resources/views.py
kofrezo: I am resetting the page to 1 in case the term changes or the DB data has changed an the user tries to access a non existent page anymore.

Let me just say that I am not too happy with the current implementation at all of this view because of a few things:
- We use Query which does not support pagination so always all objects are loaded
- Servershell search auto completion is missing here
- The method is quite big and should be split into more functions

So I will later tackle this view again but want to keep it as it is for now.

serveradmin/servershell/static/js/servershell/autocomplete/search.js
```diff
@@ -14,85 +14,111 @@ $(document).ready(function () {
...
+    let _is_attribute_value = function(to_complete) {
+        /**
+         * Check if attribute value
+         *
+         * Check if to_complete is attribute value like for example a=b and
+         * if yes return attribute and value otherwise false.
+         *
+         * @type {Array|Boolean}
+         */
+        let match = to_complete.match(/([a-z_]+)=([\S]+)?/);
```
TheFRedFox, line 26: Regexp looks good, only thing what came into my mind if it should be possible to also have something like that:
`attribute="value with spaces"`

serveradmin/servershell/static/js/servershell/autocomplete/search.js
```diff
@@ -14,85 +14,111 @@ $(document).ready(function () {
...
+
+        if (match) {
+            let attribute = match[1];
+            let value = match[2] === undefined ? '' : match[2];
+
+            return [attribute, value];
+        }
+
+        return false;
```
TheFRedFox, line 35: I would go for early return to make the actual code to be on the "root-indention":
```suggestion
        if (!match) {
            return false;
        }

        let attribute = match[1];
        let value = match[2] === undefined ? '' : match[2];

        return [attribute, value];
```

serveradmin/servershell/static/js/servershell/autocomplete/search.js
```diff
@@ -14,85 +14,111 @@ $(document).ready(function () {
...
+    };
+
+    let _complete_filter = function(to_complete) {
+        let filters = [];
+        if (to_complete) {
+            return servershell.filters.filter(filter => filter[0].startsWith(to_complete));
```
TheFRedFox, line 41: I don't know the structure of `servershell.filters.filter` right now, is `filter[0]` always safe to use, is there always a first element?

serveradmin/servershell/static/js/servershell/autocomplete/search.js
```diff
@@ -14,85 +14,111 @@ $(document).ready(function () {
...
+        }
+        else {
```
TheFRedFox, line 43: I think (not sure 100% though) you can put it in one line in js and is also code style like that:
```suggestion
        } else {
```

serveradmin/servershell/static/js/servershell/autocomplete/search.js
```diff
@@ -14,85 +14,111 @@ $(document).ready(function () {
...
             }
 
-            // Attributes available
+            // Autocomplete attributes
             let attributes = servershell.attributes.filter(
-                attr => attr.attribute_id.startsWith(cur_term) && attr.type !== 'reverse'
+                a => a.attribute_id.startsWith(to_complete) && a.type !== 'reverse'
             );
-            attributes.slice(0, limit).forEach(function (attribute) {
+            attributes.slice(0, limit).forEach(function(a) {
```
TheFRedFox, line 106: If you want it shorter, I would suggest to use `attr` as a compromise. `attr` is short and still common to read as short for `attribute`. Makes it just easier to read. If you can't go for the compromise for some reason, then ok.

serveradmin/resources/views.py
TheFRedFox: Yeah, ok, I agree. Maybe let me rephrase the first comment. An invalid `per_page` and `page` should not break or do unwanted things with the server. So, rather an awareness of checking if it is necessary to validate it or validate it early and exit early.

For the `max`, right now it would not make it any better because of the usage of `Query` for pagination, right? I would still suggest to already implement it, so it will just work when tackling the `Query` thing. As we are talking about a absolute max, we can go higher than the average of working hosts. So, I don't see anything against using 1000 as well.

serveradmin/resources/views.py
TheFRedFox: With my other comment https://github.com/innogames/serveradmin/pull/82#discussion_r402255885 feel free and do it in a way that it is just safe. ;)

serveradmin/resources/views.py
```diff
@@ -110,21 +108,36 @@ def index(request):
         for server in Query(filters, attribute_ids):
             hosts[server['hostname']] = dict(server)
 
-    page = int(request.GET.get('page', 1))
-    per_page = int(request.GET.get('per_page', 8))
-    hosts_pager = Paginator(list(hosts.values()), per_page)
+    page = abs(int(request.GET.get('page', 1)))
```
TheFRedFox, line 111: With the new comment https://github.com/innogames/serveradmin/pull/82#discussion_r402255885 feel free to just handle it in a safe way and no need to make any defaults.

serveradmin/servershell/static/js/servershell/autocomplete/search.js
kofrezo: No this is not possible and fails in a parsing error of the term.

serveradmin/servershell/static/js/servershell/autocomplete/search.js
kofrezo: Yes

serveradmin/servershell/static/js/servershell/autocomplete/search.js
kofrezo: Both is possible, I don't have any strong opinion on that so I'll take yours.

serveradmin/servershell/static/js/servershell/autocomplete/search.js
kofrezo: I just didn't like to read attribute.attribute_id and from the context I think it is visible what it is but I don't have any strong opinion on that either so I'll just revert it.

TheFRedFox: I see your point, however then the cause is the redundant `attribute` in `attribute_id`, changing the parameter name to `a` is causing that while roughly reading the code, I don't know what `a` is anymore and need to read more code to understand what it can be. `a` could stand for more things now. A developer who is used to the code would for sure directly understand it most probably and with `attr` almost every new developer will understand it right away. ;)

kofrezo: Okay I will change it to attr if you prefer that.

TheFRedFox: Yes, if we can afford the 3 extra characters (and I mean it really without sarcasm), I would prefer that.

kofrezo: Ask the clients browser ;-)

TheFRedFox: We can always minimise our js files if we are taking the client browser into account. ;)

PR #83, opened 2020-04-06
Improve error logging for api views

kofrezo:
We got reports about API calls failing with a HTTP status code 403. To be able to find the root cause this PR adds error logging for failed API calls while authentication.

### DIFF
--- a/serveradmin/api/decorators.py
+++ b/serveradmin/api/decorators.py
@@ -55,41 +55,48 @@ def api_view(view):
         then = datetime.utcfromtimestamp(
             int(request.META['HTTP_X_TIMESTAMP'])
         ).replace(tzinfo=timezone.utc)
-
-        app = authenticate_app(
-            public_keys, signatures, app_id, token, then, now, body
-        )
-
         body_json = json.loads(body) if body else None
+        status_code = 200
+
         try:
-            status_code = 200
+            app = authenticate_app(
+                public_keys, signatures, app_id, token, then, now, body
+            )
             return_value = view(request, app, body_json)
+
+            logger.info('api: Call: ' + (', '.join([
+                'Method: {}'.format(view.__name__),
+                'Application: {}'.format(app),
+                'Time elapsed: {:.3f}s'.format(
+                    (timezone.now() - now).total_seconds()
+                ),
+            ])))
         except (
             FilterValueError,
             ValidationError,
             PermissionDenied,
             ObjectDoesNotExist,
         ) as error:
+            reason = ''
+
             if isinstance(error, (FilterValueError, ValidationError)):
                 status_code = 400
+                reason = 'Bad Request'
             if isinstance(error, PermissionDenied):
                 status_code = 403
+                reason = 'Forbidden'
             if isinstance(error, ObjectDoesNotExist):
                 status_code = 404
+                reason = 'Not Found'
 
+            message = '{}: {}'.format(reason, str(error))
+            logger.error('api: {}'.format(message))
             return_value = {
                 'error': {
-                    'message': str(error),
+                    'message': message,
                 }
             }
 
-        logger.info('api: Call: ' + (', '.join([
-            'Method: {}'.format(view.__name__),
-            'Application: {}'.format(app),
-            'Time elapsed: {:.3f}s'.format(
-                (timezone.now() - now).total_seconds()
-            ),
-        ])))
         return HttpResponse(
             json.dumps(return_value, default=json_encode_extra),
             content_type='application/x-json',


serveradmin/api/decorators.py
```diff
@@ -55,15 +55,22 @@ def _wrapper(request):
...
-        )
-
         body_json = json.loads(body) if body else None
+        status_code = 200
+
         try:
-            status_code = 200
+            app = authenticate_app(
+                public_keys, signatures, app_id, token, then, now, body
+            )
```
HerrSpace, line 64: So here's what this will do:
1) make sure we return a proper JSON API response on auth errors instead of the default django error page
2) tell the client _why_ the authentication was forbidden.

I'm not quite sure about the security implications of 2, though I think it should be fine. 1 is definitely a good thing!

serveradmin/api/decorators.py
```diff
@@ -77,19 +84,22 @@ def _wrapper(request):
...
                 }
             }
 
-        logger.info('api: Call: ' + (', '.join([
-            'Method: {}'.format(view.__name__),
-            'Application: {}'.format(app),
-            'Time elapsed: {:.3f}s'.format(
-                (timezone.now() - now).total_seconds()
-            ),
-        ])))
```
HerrSpace, line 92: This can stay where it is if you just log the error above.

HerrSpace:
If you grep for `raise PermissionDenied` you get all the places 403s can come from (only 14 lines!). I'll just assume it's not logintool deactivating users or ACLs misbehaving as those things don't happen intermittently.

Did you recently have NTP issues? The only thing I can see is adminapis replay defense: https://github.com/innogames/serveradmin/pull/62#discussion_r273937154 As you can see we even lowered the threshold a year ago.

kofrezo:
> While I think moving authenticate_app into the try block might be good idea, this whole PR has nothing todo with what's in the headline. This will not fix random 403s.

Hi Patrick,
Thanks for your feedback. Actually it does. Because tough I can not explain why because I do not understand nor do two other people I asked about. The errors are gone with this change in a reproducible way. As soon as I revert the change and run a few thousand requests it happens again.

If you can explain or have a hint for me or have a suggestion for a better title which includes both I would appreciate that.

serveradmin/api/decorators.py
```diff
@@ -77,19 +84,22 @@ def _wrapper(request):
...
+            logger.error('{} request data: {}'.format(str(error), {
+                'now': now,
+                'body': body,
+                'public_keys': public_keys,
+                'signatures': signatures,
+                'app_id': app_id,
+                'token': token,
+                'then': then,
+                'body_json': body_json,
+            }), exc_info=True)
```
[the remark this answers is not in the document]
kofrezo: > Things logged here aren't serveradmin errors but client errors
I am pretty sure we have a backend error here as we can reproduce the error by changing the backend code.

> Also this will likely log a huge amount of data when things go south and that data will not be readily understandable by humans either
I disagree with that. The change actually helped a lot in debugging the error and making sure there was no wrong request data sent to the backend.

> If any you should bring back something like logger.error('api: Forbidden: ' + problem) here.
I think we are almost there. How about we take your suggestion to log the message as you suggested and log the request data on level debug which allows the person running the servers to log it to a custom location if needed?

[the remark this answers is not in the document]
kofrezo: > we explicitly switched away from this approach to raising Exceptions instead

Do you have any details why we decided to go for this approach. The problem I have here is that I get bug reports from users not getting to much details about why the 403 happened and I have no way to prove that it is client side.

serveradmin/api/decorators.py
kofrezo: Sorry I am confused. Do you consider this now correct or no ?

serveradmin/api/decorators.py
kofrezo: Correct me if I am wrong but I don't thinks so because it would potentially cause a NameError in python because authenticate_app might fail without declaring and setting the app variable which would then fail here.

serveradmin/api/decorators.py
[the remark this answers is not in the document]
kofrezo: To give you a bit more context about what we are talking. This is a exception we are facing without the change but not anymore with the change (I removed the data part as it contains sensitive data but I was able to redo the request with the request data successfully):

`
[2020-03-27 20:23:27] serveradmin._wrapper:97 ERROR Disabled application request data: [...]
  File "/usr/lib/python3/dist-packages/serveradmin/api/decorators.py", line 64, in _wrapper
    public_keys, signatures, app_id, token, then, now, body
  File "/usr/lib/python3/dist-packages/serveradmin/api/decorators.py", line 151, in authenticate_app
    raise PermissionDenied('Disabled application')
django.core.exceptions.PermissionDenied: Disabled application
`

It raises the exception here: https://github.com/innogames/serveradmin/blob/master/serveradmin/api/decorators.py#L140

but it is completely dubious to me how. I even installed a trigger on the database and no application was disabled at any time while this error occurred. Unfortunately I don't see how the changes related can this to be fixed.

serveradmin/api/decorators.py
HerrSpace: True, sorry.

serveradmin/api/decorators.py
HerrSpace: I think it's probably a good idea, as discussed in the call.

HerrSpace:
Call wrapup:

In this branch:
- Add SuspiciousOperation to except block
- Remove the logger call above (logs psks, logs lots of data, possible DOS)
- Add logger call like `logger.error('api: Forbidden: ' + problem)`

In a new branched forked from master:
- Modify 'Inactive user' and 'Disabled application' Exceptions to include user and app ids to see whether we might get to this stack trace by sometimes returning the wrong application during logins.
- Add logger call like `logger.error('api: Forbidden: ' + problem)`
- Try to reproduce the error with this branch and see if the logged app_id is actually disabled on serveradmin.

kofrezo:
After a call with @HerrSpace we agreed on the following:

- [ ] Fix miss leading PR summary: Make clear that we just extend the error logging
- [ ] Do not log request data as it contains sensitive data such as auth tokens
- [ ] Log errors and reason @HerrSpace suggested: logger.error('api: Forbidden: ' + problem)
- [ ] Capture missing SuspiciousOperation exception
- [ ] Tell client why the request failed (ok from security perspective)
- [ ] Log user id and application if user or app is disabled

We can not explain the reason for the disabled application exception shown above but as it is potentially something dangerous such as wrong loaded application we will further investigate this and try to reproduce.

serveradmin/api/decorators.py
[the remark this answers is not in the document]
kofrezo: See https://github.com/innogames/serveradmin/pull/83#issuecomment-609846007

serveradmin/api/decorators.py
kofrezo: This turns out to be just a summary/explanation of the desired changes.

kofrezo:
> If you grep for `raise PermissionDenied` you get all the places 403s can come from (only 14 lines!). I'll just assume it's not logintool deactivating users or ACLs misbehaving as those things don't happen intermittently.
> 
> Did you recently have NTP issues? The only thing I can see is adminapis replay defense: [#62 (comment)](https://github.com/innogames/serveradmin/pull/62#discussion_r273937154) As you can see we even lowered the threshold a year ago.

- I grepped for PermissionDenied but there is only one occurence with 'PermissionDenied('Disabled application' as we see it in the logs (https://github.com/innogames/serveradmin/blob/adminapi_403/serveradmin/api/decorators.py#L150)
-  We have indeed problems with expired requests but they are logged with 'Request expired' and we see them also with this message (https://github.com/innogames/serveradmin/blob/master/serveradmin/api/decorators.py#L123)

They were not distinguishable before (both 403 due to permission denied) but we see them now as two exceptions.

PR #84, opened 2020-04-06
task(common): Delete unused templatetags

kofrezo:


kofrezo:
Both logo and gethostname were used in the old design but are no longer.

PR #85, opened 2020-04-07
task(api): Add user or application id to error

kofrezo:
Add user and application id to PermissionDenied error to be able to see which user/application was disabled in error log. As the error message is also send to the client as response we only use the user/application id to not expose sensitive information.

### DIFF
--- a/serveradmin/api/decorators.py
+++ b/serveradmin/api/decorators.py
@@ -134,10 +134,10 @@ def authenticate_app(
         raise SuspiciousOperation('Missing authentication')
 
     if app.owner is not None and not app.owner.is_active:
-        raise PermissionDenied('Inactive user')
+        raise PermissionDenied('Inactive user {}'.format(app.owner.id))
 
     if app.disabled:
-        raise PermissionDenied('Disabled application')
+        raise PermissionDenied('Disabled application {}'.format(app.id))
 
     # Note when this app was last used. We don't update this information more
     # than once every minute. Some apps authenticate thousand of times per


serveradmin/api/decorators.py
```diff
@@ -134,10 +134,10 @@ def authenticate_app(
         raise SuspiciousOperation('Missing authentication')
 
     if app.owner is not None and not app.owner.is_active:
-        raise PermissionDenied('Inactive user')
+        raise PermissionDenied('Inactive user {}'.format(app.owner.id))
```
TheFRedFox, line 137: Just to clarify:
I was thinking it through for Enumeration attacks. I am not sure. Is it possible to get to an app e.g. with `app_id` and `token` and ending up with an `owner` not being the requester, which could be used for existence enumeration?

kofrezo: Just for clarification: The app_id is the result of the sha1 sum of the auth token (secret). The token is the result of the hmac of payload + timestamp. 

This means you would have to guess a correct app_id (secret) with a disabled user or application to get any of this message.

You can already find out whatever a application exists or not already: 

https://github.com/innogames/serveradmin/blob/master/serveradmin/api/decorators.py#L166

From security perspective I think it is already hard enough to find out the secret - no ?

kofrezo: I looked it up. According to our cryptography standards which follow the trusted standards SHA1 + HMAC is considered to be secure enough tough not recommended anymore for new applications. As this is not touched by this PR I consider it safe and will not further investigate it.

PR #88, opened 2020-04-11
fix(servershell): Use ServertypeAttribute default visible

kofrezo:
The state attribute is InnoGames internal and we have a default visible
attribute for each Servertype we can use to generate a default list of
attributes visible in Servershell.

### DIFF
--- a/serveradmin/servershell/helper/__init__.py
+++ b/serveradmin/servershell/helper/__init__.py
@@ -1,4 +1,4 @@
-from serveradmin.serverdb.models import Attribute
+from serveradmin.serverdb.models import Attribute, ServertypeAttribute
 
 
 def get_default_shown_attributes():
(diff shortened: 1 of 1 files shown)


serveradmin/servershell/helper/__init__.py
```diff
@@ -9,7 +9,8 @@ def get_default_shown_attributes():
 
     shown_attributes = list(Attribute.specials.keys())
     shown_attributes.remove('object_id')
-    shown_attributes.append('state')
+    shown_attributes.extend([attr.attribute_id for attr in
+                             ServertypeAttribute.objects.filter(default_visible=True).only('attribute_id').distinct()])
```
TheFRedFox, line 13: To have a better readability I would suggest to store this line in an own variable like `default_attributes` or the like.

PR #89, opened 2020-04-14
feature(adminapi): Send version with each request

kofrezo:


adminapi/request.py
```diff
@@ -137,6 +140,7 @@ def _build_request(endpoint, get_params, post_params):
     headers = {
         'Content-Encoding': 'application/x-json',
         'X-Timestamp': str(timestamp),
+        'X-API-Version': '.'.join(str(v) for v in VERSION),
```
TheFRedFox, line 143: What's the plan for using that? Do I get it correctly it is to identify the versions of the clients? The `User-Agent` header could then also be a possible header to use with like `python 3.42; adminapi 42.42.42` or the like.

kofrezo: It is useful to identify servers running old version of the API. In the current case I still see some errors but I am not sure whatever the client had already the latest version or not. I can of course determine the IP and SSH to the machine and check but for some servers like SNAT(ed) ones that is impossible.

Generally all servers should be configured to automatically update to the latest adminapi version when running puppet for us but for if you use Serveradmin somewhere else or have disabled puppet or disabled automatic upgrading this is helpful.

I want to keep User Agent as it helps me to understand what Python version/urllib version is running. Having both can be useful.

TheFRedFox: Yep, understood. I was also rather thinking of extending the `User-Agent` than replacing it completely. I guess that might take some more time to figure out how it can be done nicely and it might be interesting for the future to have an `User-Agent` like: `python/3.42 urllib/42 adminapi/42.42.42` or however/whatever you want it to be.

kofrezo: Do you see any benefit from that ? I like having it a dedicated header to compare it more easily without string parsing.

TheFRedFox: `User-Agent` sounds just the way to go there, as it is actually the user agent. A dedicated header still might make sense if you need something more faster and straightforward. To get a complete matrix of version combination: python | urllib | adminapi , the `User-Agent` seems more appropriate than having dedicated headers for each one. Still not saying that dedicated headers should not be used at all. I just see dedicated headers as an extension of the `User-Agent`, so to go both ways in the future might make sense.

PR #90, opened 2020-04-14
task: Update distribute and python versions for tests

kofrezo:
- Use Ubuntu Bionic 18.04 LTS for running tests
- Test all needed Python versions 3.5 (Debian Stretch) to 3.7 (Debian Buster)

This drops support for Python 3.4. Support for Python 3.8 is not yet in tests break for Django 1.10 we have to upgrade Django first.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -13,7 +13,7 @@ formats: all
 
 # Optionally set the version of Python and requirements required to build your docs
 python:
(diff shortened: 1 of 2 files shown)


kofrezo:
Debian Jessie is still officially supported and we still have a few machines running but we also have many buster machines running Python 3.7. I will therefore try to find a way to test all python versions.

PR #91, opened 2020-04-29
fix(servershell): Cannot submit multi values of int and bool

kofrezo:
- Add cast to int and bool for multi attributes
- Fix filter for empty values before updating attribute
- Fix cast values before comparing in inline editing

### DIFF
--- a/serveradmin/servershell/static/js/servershell/command.js
+++ b/serveradmin/servershell/static/js/servershell/command.js
@@ -1,4 +1,25 @@
 /**
+ * Transform value into matching type
+ *
+ * We get user input always as string. This method transforms values for
+ * number, boolean etc. attributes in the matching type
+ *
+ * @param value
+ * @param attribute
+ * @returns {*}
+ */
+servershell.transform_value = function(value, attribute) {
+    if (attribute.type === 'number') {
+        return attribute.multi ? value.map(v => Number.parseInt(v)) : Number.parseInt(value);
+    }
+    else if (attribute.type === 'boolean') {
+        return attribute.multi ? value.map(v => !!v) : !!value;
+    }
+
+    return value;
+}
+
+/**
  * Update (multi) attribute values
  *
  * Adds or removes values for the given object no matter if it is a multi
@@ -21,8 +42,7 @@ servershell.update_attribute = function(object_id, attribute_id, new_value, mult
     let server = servershell.get_object(object_id);
 
     // Cast new values to corresponding data type if wanted
-    new_value = attribute.type === 'number' ? Number.parseInt(new_value) : new_value;
-    new_value = attribute.type === 'boolean' ? !!new_value : new_value;
+    new_value = servershell.transform_value(new_value, attribute);
 
     // No change for this object about to commit yet
     if (!changes.hasOwnProperty(object_id)) {
@@ -62,7 +82,9 @@ servershell.update_attribute = function(object_id, attribute_id, new_value, mult
         let to_remove;
 
         // Don't add or remove empty values
-        new_value = new_value.filter(v => v.trim() !== '');
+        new_value = new_value.filter(function(v) {
+            return (v + '').trim() !== '';
+        });
 
         // Merge changes from previous commands not committed yet
         if (multi_action === 'add') {
--- a/serveradmin/servershell/static/js/servershell/result.js
+++ b/serveradmin/servershell/static/js/servershell/result.js
@@ -244,14 +244,17 @@ register_inline_editing = function(cell) {
             let attribute = servershell.get_attribute(attribute_id);
 
             // When the user types 'false' use empty string so that it casts to false
-            if (attribute.type === 'boolean' && (value === 'false' || value === '0'))
+            if (attribute.type === 'boolean' && (value === 'false' || value === '0')) {
                 value = '';
+            }
 
             if (value === '') {
                 servershell.delete_attribute(object_id, attribute_id)
             }
             else {
                 if (multi) {
+                    value = servershell.transform_value(value, attribute);
+
                     let current_value = servershell.get_object(object_id)[attribute_id];
                     let to_add = value.filter(v => !current_value.includes(v));
                     let to_remove = current_value.filter(v => !value.includes(v));


serveradmin/servershell/static/js/servershell/command.js
```diff
@@ -62,7 +76,9 @@ servershell.update_attribute = function(object_id, attribute_id, new_value, mult
         let to_remove;
 
         // Don't add or remove empty values
-        new_value = new_value.filter(v => v.trim() !== '');
+        new_value = new_value.filter(function(v) {
```
TheFRedFox, line 79: What was the reason to change the syntax from the arrow-function syntax?

serveradmin/servershell/static/js/servershell/command.js
```diff
@@ -62,7 +76,9 @@ servershell.update_attribute = function(object_id, attribute_id, new_value, mult
...
+            return (v + '').trim() !== '';
```
TheFRedFox, line 80: What makes it needed to add `''` to force it to a string and not e.g. using `String(v)`?

serveradmin/servershell/static/js/servershell/result.js
```diff
@@ -244,14 +244,26 @@ register_inline_editing = function(cell) {
...
+            }
 
             if (value === '') {
                 servershell.delete_attribute(object_id, attribute_id)
             }
             else {
                 if (multi) {
+                    // Cast to matching type before comparing
+                    value = value.map(function(v) {
+                       if (attribute.type === 'number') {
```
TheFRedFox, line 258: Above it looks like a multi value can just have values of one attribute type. So, multiple numbers or multiple booleans or the like. Now it looks like a multi value can have values of different attribute types.
What is correct? As I would suggest to rather use one way of handling it, if this is possible. Then you can easily modulize it later.

serveradmin/servershell/static/js/servershell/command.js
```diff
@@ -21,8 +21,22 @@ servershell.update_attribute = function(object_id, attribute_id, new_value, mult
     let server = servershell.get_object(object_id);
 
     // Cast new values to corresponding data type if wanted
-    new_value = attribute.type === 'number' ? Number.parseInt(new_value) : new_value;
-    new_value = attribute.type === 'boolean' ? !!new_value : new_value;
+    if (attribute.type === 'number') {
+        if (attribute.multi) {
+            new_value = new_value.map(v => Number.parseInt(v));
```
TheFRedFox, line 26: What about if `parseInt` returns `NaN`? Is this safe to handle?

TheFRedFox, line 26: What about floats? Is `'number'` attribute type always just `int`s?

kofrezo: If the user enter 'a' into a number field and ignored the red background due to regex warning it is supposed to fail here yes.

kofrezo: Yes we don't support floats right now.

serveradmin/servershell/static/js/servershell/command.js
kofrezo: None

serveradmin/servershell/static/js/servershell/command.js
kofrezo: Yes. What is the benfit of String(v) over ' ' + v ?

serveradmin/servershell/static/js/servershell/result.js
kofrezo: A multi attribute can only have values of one type. I just wrote it this way because it is shorter.

serveradmin/servershell/static/js/servershell/command.js
TheFRedFox: If I see it correctly, `Number.parseInt('a')` won't fail, it would return a `NaN` number object.
image

kofrezo: It will fail at latest on the commit to the backend. Adding another part validating here would duplicate the logic imho. Again if you type a 'a' into a field where a number is expected it should jump into your face already before saving

Screenshot from 2020-04-29 15-47-56

at least after saving and before commiting:

Screenshot from 2020-04-29 15-48-04

serveradmin/servershell/static/js/servershell/command.js
TheFRedFox: Not much actually. `String(v)` looks more elegant to me. Read also this https://stackoverflow.com/a/3945236/2625488 and experimented a bit:
image

serveradmin/servershell/static/js/servershell/command.js
kofrezo: I can add a manual extra check here but this would then be arguable the same for boolean and all other attribute types we currently have and will add in future.

Keeping the validation work in sync between backend and frontend would be tedious. For example we have a type datetime.

TheFRedFox: Question was if it is safe to handle. If at latest on backend side it will fail, then I would consider it safe enough.

serveradmin/servershell/static/js/servershell/result.js
TheFRedFox: Then I would suggest to use one way, to easier modularize it later. The way in this line might be a bit slower as you are doing not needed checks for the attribute type for every value, even though the attribute type can't be different.

serveradmin/servershell/static/js/servershell/command.js
kofrezo: Thanks for the research. I did not knew that. 

All tough passing non primitive types is not supported by Servershell so it should not be a use case, I think `v = ''` is the better choice because if we ever come to this point we want to have the primitive representation of valueOf() and not the String representation of that object.

serveradmin/servershell/static/js/servershell/command.js
kofrezo: Agreed

serveradmin/servershell/static/js/servershell/result.js
kofrezo: Modularisation is not planned in the near future. Since we only put in small parts of texts (maybe max 256 characters) I would consider that a micro optimisation.

I will adjust it.

PR #93, opened 2020-05-18
Add unit tests for command line interface

kofrezo:
Add unit tests for the argument parser of the command line interface to guarantee a stable interface not breaking clients or users behavior on changes.

### DIFF
--- a/adminapi/tests/test_cli.py
+++ b/adminapi/tests/test_cli.py
@@ -1,0 +1,68 @@
+import unittest
+from argparse import ArgumentParser
+from typing import Text, NoReturn
+
+from adminapi.cli import parse_args
+
+
+# ArgumentParser exists on error, we cannot capture this so we override it
+class ArgumentParserMock(ArgumentParser):
+    def error(self, message: Text) -> NoReturn:
+        raise Exception(message)
+
+
+class TestCommandlineInterface(unittest.TestCase):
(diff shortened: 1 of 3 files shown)


adminapi/tests/test_cli.py
```diff
@@ -0,0 +1,74 @@
...
+from typing import Text, NoReturn
+from unittest import mock
+
+from adminapi.cli import parse_args
+
+
+# ArgumentParser exists on error, we cannot capture this so we override it
+class ArgumentParserMock(ArgumentParser):
+    def error(self, message: Text) -> NoReturn:
+        raise Exception(message)
```
TheFRedFox, line 12: Valid solution. Only question would be, if using a less broader Exception such like `ValueError` would also be possible.

And also to check out another possible solution: https://stackoverflow.com/a/53675033/2625488 as this would have been also my first approach to test it.

kofrezo: I thought also about using ValueError as described in the StackOverflow but it is not really a ValueError and I did not see any benefit in creating a custom Exception type just for unit tests all tough I of course see the necessity in using specific Exceptions in the normal program workflow.

I think there is value from asserting the SystemExit as it eliminates the need of extra code for mocking and is closer to the real behaviour - I will use that.

PR #94, opened 2020-05-19
fix: Support automatic url inclusion for AppConfig

kofrezo:
We provide a feature which automatically includes the url configs of
apps if they are in the serveradmin package or start with serveradmin_
but this does not work for apps using the AppConfig.

This supports both the classic importing the whole package and the
AppConfig.

### DIFF
--- a/serveradmin/urls.py
+++ b/serveradmin/urls.py
@@ -1,10 +1,11 @@
 """Serveradmin
 
-Copyright (c) 2019 InnoGames GmbH
+Copyright (c) 2020 InnoGames GmbH
 """
 
 from importlib.util import find_spec
 
+from django.apps import apps
 from django.conf import settings
 from django.conf.urls import include, url
 from django.conf.urls.static import static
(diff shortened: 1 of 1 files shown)


serveradmin/urls.py
```diff
@@ -25,6 +25,9 @@
 ]
 
 for app in settings.INSTALLED_APPS:
+    if '.apps.' in app and app.endswith('Config'):
```
TheFRedFox, line 28: The names are unfortunately just convention and can be completely different. If this is fine for us, then ok.
With `apps.get_app_configs()` (https://docs.djangoproject.com/en/1.10/ref/applications/#django.apps.apps.get_app_configs) we could have a more reliable solution though.

kofrezo: Neat! Didn't knew that will do.

serveradmin/urls.py
```diff
@@ -24,16 +25,17 @@
     url(r'^admin/', admin.site.urls),
 ]
 
-for app in settings.INSTALLED_APPS:
-    module_spec = find_spec(app + '.urls')
+for app in apps.get_app_configs():
+    name = app.name
```
TheFRedFox, line 29: This should definitely already work and is the pendant to the previous implementation.
I wonder if we can also make use of the module property of the AppConfigs: https://docs.djangoproject.com/en/1.10/ref/applications/#django.apps.AppConfig.module
Just a mentioning here. Maybe it would be even more reliable in the future.

kofrezo: Do you have anything specific in mind ? 

We try to load the urls module which is part of the app module. This might or might not exist - both is valid. I don't see how the app module could help in here or how we can access the urls module from the app module in a way better than this.

PR #95, opened 2020-05-25
fix: Skip app if urls module does not exist

kofrezo:


serveradmin/urls.py
```diff
@@ -28,13 +28,14 @@
 for app in apps.get_app_configs():
     name = app.name
     module_spec = find_spec(name + '.urls')
-    if module_spec is not None:
-        module = module_spec.loader.load_module()
+    if module_spec is None:
+        continue
```
TheFRedFox, line 32: Yes, absolutely, sry I didn't see that in the last one already.

PR #96, opened 2020-06-02
serveradmin(resources): add management command for migration log cleanup

konstantinbest:
This command will check the igvm_migration_log of every Hypervisor 
and discard all entries which are not from the same day. We can use 
this within a Cron job.

### DIFF
--- a/serveradmin/resources/management/commands/clear_migration_log.py
+++ b/serveradmin/resources/management/commands/clear_migration_log.py
@@ -1,0 +1,49 @@
+"""Serveradmin - igvm_migration_log cleanup
+
+We are using igvm_migration_log for calculating the overall CPU usage of a
+Hypervisor. This is obsolete after the next prime time because the cpu_util_pct
+will be correct again. This task will get executed once in the morning to
+clean up the log entries from the previous day.
+
+Copyright (c) 2020 InnoGames GmbH
+"""
+from datetime import datetime, timedelta, timezone
+from django.core.management.base import BaseCommand
+
+from adminapi.dataset import Query
+from adminapi.filters import Not, Empty
+
+
+class Command(BaseCommand):
+    help = 'Deletes igvm_migration_log entries from previous day'
+
+    def handle(self, *args, **options):
+        """Entry point for command
+
+        get all hypervisors which have an entry in igvm_migration_log
+        and delete all entries which are from the previous day.
+        """
+
+        all_hypervisors = Query({'igvm_migration_log': Not(Empty()),
+                                 'servertype': 'hypervisor'},
+                           ['igvm_migration_log', 'hostname'])
+
+        now = datetime.now(tz=timezone.utc)
+
+        for hypervisor in all_hypervisors:
+
+            for entry in hypervisor['igvm_migration_log']:
+                timestamp = entry.split()[0]
+
+                dt_object_sa = datetime.fromtimestamp(
+                    int(timestamp),
+                    tz=timezone.utc)
+
+                if dt_object_sa.date() == now.date():
+                    continue
+
+                hypervisor['igvm_migration_log'].discard(entry)
+                hypervisor.commit()
+                self.stdout.write(self.style.SUCCESS(
+                    'Delete old entries from {} '.format(
+                        hypervisor['hostname'])))


serveradmin/resources/management/commands/clear_migration_log.py
```diff
@@ -0,0 +1,49 @@
+"""Serveradmin - igvm_migration_log cleanup
+
```
lacunoc, line 2: Please add a short summary about what the command does and why we need this.

serveradmin/resources/management/commands/clear_migration_log.py
```diff
@@ -0,0 +1,49 @@
...
+Copyright (c) 2020 InnoGames GmbH
+"""
+from django.core.management.base import BaseCommand
```
lacunoc, line 5: Let's order them by alphabetical first and put the IG ones below.

serveradmin/resources/management/commands/clear_migration_log.py
```diff
@@ -0,0 +1,49 @@
...
+from adminapi.dataset import Query
+from adminapi.filters import Not, Empty
+from datetime import datetime
+
+class Command(BaseCommand):
+    help = 'Check for old igvm_migration_log entries and deletes them'
```
lacunoc, line 11: typo
`Checks...`

serveradmin/resources/management/commands/clear_migration_log.py
```diff
@@ -0,0 +1,49 @@
...
+
+        get all hypervisors which have an entry in igvm_migration_log
+        and delete all entries which are from the previous day.
+
+        """
+
+        all_hypervisor = Query({'igvm_migration_log': Not(Empty())},
+                           ['igvm_migration_log', 'hostname'])
+
+        for hypervisor in all_hypervisor:
```
TuxPowered42, line 24: Shouldn't each HV be `igvm_locked` for this operation?

serveradmin/resources/management/commands/clear_migration_log.py
```diff
@@ -0,0 +1,49 @@
...
+
```
lacunoc, line 25: Not needed empty line here.

serveradmin/resources/management/commands/clear_migration_log.py
```diff
@@ -0,0 +1,49 @@
...
+    def handle(self, *args, **options):
+        """Entry point for command
+
+        get all hypervisors which have an entry in igvm_migration_log
+        and delete all entries which are from the previous day.
+
+        """
+
+        all_hypervisor = Query({'igvm_migration_log': Not(Empty())},
```
lacunoc, line 21: Does it make any difference performance wise to also filter this for `'servertype': 'hypervisor'` only?

serveradmin/resources/management/commands/clear_migration_log.py
```diff
@@ -0,0 +1,49 @@
...
+
+                dt_object_sa = datetime.fromtimestamp(float(timestamp))
+                day_sa = dt_object_sa.day
+                month_sa = dt_object_sa.month
+
+                current_month = datetime.now().month
+                current_day = datetime.now().day
+
+                if (day_sa - current_day == 0 and
+                        current_month - month_sa == 0):
```
lacunoc, line 37: What about `year`?

serveradmin/resources/management/commands/clear_migration_log.py
```diff
@@ -0,0 +1,49 @@
...
+
+        """
+
+        all_hypervisor = Query({'igvm_migration_log': Not(Empty())},
+                           ['igvm_migration_log', 'hostname'])
+
+        for hypervisor in all_hypervisor:
+
+            for entry in hypervisor['igvm_migration_log']:
+                timestamp = entry.split(' ')[0]
```
TheFRedFox, line 27: `split(' ')` will split on exactly ONE space. On default split will split like that:
```
split(sep=None, maxsplit=-1) method of builtins.str instance
    Return a list of the words in the string, using sep as the delimiter string.

    sep
      The delimiter according which to split the string.
      None (the default value) means split according to any whitespace,
      and discard empty strings from the result.
    maxsplit
      Maximum number of splits to do.
      -1 (the default value) means no limit.
```

To show it in action on an example:
```
>>> a = 'hello   h'
>>> a.split()
['hello', 'h']
>>> a.split(' ')
['hello', '', '', 'h']
>>>
```

The question is therefore, what's important to have explicitly `' '` and not the default behavior of split?

serveradmin/resources/management/commands/clear_migration_log.py
```diff
@@ -0,0 +1,49 @@
...
+
+                dt_object_sa = datetime.fromtimestamp(float(timestamp))
```
kofrezo, line 29: integer is sufficient here and makes more sense as we also store the values as integer in the attribute value.

serveradmin/resources/management/commands/clear_migration_log.py
kofrezo: The timedelta class could be useful here: https://docs.python.org/3/library/datetime.html#timedelta-objects

serveradmin/resources/management/commands/clear_migration_log.py
konstantinbest: Why should it? The cron itself will run once in the morning and we don't interfere with anything during this cleanup.

serveradmin/resources/management/commands/clear_migration_log.py
konstantinbest: Since the entry has excact one space, we can use the default behavior.

serveradmin/resources/management/commands/clear_migration_log.py
konstantinbest: integrated timedelta class.

serveradmin/resources/management/commands/clear_migration_log.py
```diff
@@ -0,0 +1,49 @@
...
+                day_sa = dt_object_sa.day
+                month_sa = dt_object_sa.month
+
+                current_month = datetime.now().month
+                current_day = datetime.now().day
+
+                if (day_sa - current_day == 0 and
+                        current_month - month_sa == 0):
+                    continue
```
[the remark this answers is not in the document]
konstantinbest: No problem anymore after implementing the timedelta class.

serveradmin/resources/management/commands/clear_migration_log.py
konstantinbest: added the servertype.

serveradmin/resources/management/commands/clear_migration_log.py
```diff
@@ -0,0 +1,50 @@
...
+
+        get all hypervisors which have an entry in igvm_migration_log
+        and delete all entries which are from the previous day.
+        """
+
+        all_hypervisors = Query({'igvm_migration_log': Not(Empty()),
+                                 'servertype': 'hypervisor'},
+                           ['igvm_migration_log', 'hostname'])
+
+        current_timestamp = datetime.utcnow()
```
TheFRedFox, line 31: Don't use `utcnow` as it will behave weirdly than what you would expect in combination with timezone aware datetime objects, as without a timezone the datetime object will be treated as local time. @Felixoid was giving a lightning talk about that topic. Try to always use timezone aware datetime objects with `now` and even better with a timezone. You can specify the timezone as utc with that:
```suggestion
        current_timestamp = datetime.now(tz=timezone.utc)
```

TheFRedFox, line 31: How about just `now` instead of `current_timestamp`, especially as it is not a `timestamp` but a datetime object?
```suggestion
        now = datetime.utcnow()
```

serveradmin/resources/management/commands/clear_migration_log.py
```diff
@@ -0,0 +1,50 @@
...
+        hour = current_timestamp.hour
+
+        for hypervisor in all_hypervisors:
+
+            for entry in hypervisor['igvm_migration_log']:
+                timestamp = entry.split()[0]
+
+                dt_object_sa = datetime.fromtimestamp(int(timestamp))
```
TheFRedFox, line 39: When working with dates and times, try to be consistent with the timezones. `fromtimestamp` without the timezone parameter will convert the timestamp to the local time without being timezone aware.
With `utcnow` you would now compare a timezone-naive (which will be treated as the local time, although it is utc time) with a timezone-naive local datetime. If the local time of the system is not also utc, this will behave wrongly.
With `now` (without specifying the utc timezone) you would now compare a timezone-naive local time with a timezone-naive conversion of a timestamp to local time, which should behave correctly
With `now` and utc timezone, you would now compare a timezone-aware utc time with a timezone-naive conversion of a timestamp to local time, which is not even possible.
With `now` and utc timezone and `fromtimestamp(timestamp, tz=timezone.utc)` you would now compare a timezone-aware utc time with a timezone-aware conversion of a timestamp to utc time, which is behaving correctly.

serveradmin/resources/management/commands/clear_migration_log.py
TuxPowered42: If cron runs during a migration, igvm won't be able to commit data with Serveradmin error "newer data available". Please don't write opportunistic software.

kofrezo: This is only true for single attributes @TuxPowered42. For multi attributes this is not the case - so it would not fail.

```
>>> cron = Query({'hostname': 'somehostl'}, ['hostname', 'igvm_migration_log', 'backup_disabled']).get()
>>> igvm = Query({'hostname': 'somehost'}, ['hostname', 'igvm_migration_log', 'backup_disabled']).get()
>>> cron['igvm_migration_log'].add('1591179865 +5')
>>> igvm['igvm_migration_log'].add('1591179865 -5')
>>> cron.commit()
>>> igvm.commit()
>>> 
>>> cron['backup_disabled'] = False
>>> igvm['backup_disabled'] = False
>>> cron.commit()
>>> igvm.commit()
  File "<console>", line 1, in <module>
  File "/home/kofrezo/Sourcecode/serveradmin/adminapi/dataset/__init__.py", line 456, in commit
    _handle_exception(result)
  File "/home/kofrezo/Sourcecode/serveradmin/adminapi/dataset/__init__.py", line 542, in _handle_exception
    raise exception_class(result['message'])
adminapi.exceptions.DatasetError: ['Newer data available']
```

In this specific case there is also no logical conflict possible from what I understood because igvm will only add new values and the cron job will only remove values.

TuxPowered42: In this case it's fine.

PR #98, opened 2020-06-03
fix(clear_migration_log): add missing __init__ files

konstantinbest:
Add missing __init__.py files for new management command.


kofrezo:
> Add missing __init__.py files for new management command.
While Django can still find the files even when they are not python modules the packaging job only contains python modules. Thus said this is necessary.

PR #103, opened 2020-06-23
fix: Move VERSION from serveradmin to adminapi

kofrezo:
Running serveradmin without adminapi module is impossible right now due
to code dependencies. Running adminapi without serveradmin however
should be possible but became impossible by sending the version as
header.

### DIFF
--- a/adminapi/__init__.py
+++ b/adminapi/__init__.py
@@ -1,17 +1,4 @@
 """Serveradmin - adminapi
 
-Copyright (c) 2018 InnoGames GmbH
+Copyright (c) 2020 InnoGames GmbH
 """
-
-from adminapi.request import Settings
-
-
-# XXX Deprecated
-def auth(auth_token=None):
-    if auth_token:
-        Settings.auth_token = auth_token
-
-
-# XXX Deprecated
-def set_timeout(timeout, what='api'):
-    Settings.timeout = timeout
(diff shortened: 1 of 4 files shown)


kofrezo:
This makes it possible to package adminapi module without serveradmin depdency currently used for Arch Linux.

serveradmin/__init__.py
```diff
@@ -2,5 +2,3 @@
 
 Copyright (c) 2020 InnoGames GmbH
 """
-
-VERSION = (1, 9, 3)
```
Felixoid, line 6: Shouldn't it be `from adminapi import VERSION`?

kofrezo: This would be possible too. I decided to adjust all current imports so that there is only one origin to check and one does not have to go along a chain of imports.

adminapi/__init__.py
```diff
@@ -1,17 +1,6 @@
...
-# XXX Deprecated
-def auth(auth_token=None):
-    if auth_token:
-        Settings.auth_token = auth_token
-
-
-# XXX Deprecated
-def set_timeout(timeout, what='api'):
-    Settings.timeout = timeout
+VERSION = (1, 9, 3)
```
Felixoid, line 6: Is it possible to bump version here as well, please?

kofrezo: I increased the version by accident. This is fixed. This should be the only place where we have to bump the version for now.

PR #105, opened 2020-07-06
Make Serveradmin Django 2.2 compatible

kofrezo:
- Use new syntax for urls.py
- Fix LANGUAGE_CODE setting
- Remove obsolete function from view
- Remove obsolete interactive module from adminapi
- Remove obsolete export view


Pipfile
```diff
@@ -15,7 +15,7 @@ netaddr = "*"
 # WSGI server
 gunicorn = "*"
 # Debian stretch version
-Django = "~=1.11.0"
+Django = "==2.0.13"
```
TheFRedFox: The version doesn't fit the one in the commit message. ^^

kofrezo: True. I must have mixed up minor version here. However since it is later on upgraded to 2.2.13 and tested against this it should be fine.

PR #112, opened 2020-10-06
Remove deprecated dataset_create function

kofrezo:


adminapi/dataset/__init__.py
```diff
@@ -16,7 +16,6 @@
 NEW_OBJECT_ENDPOINT = '/dataset/new_object'
 COMMIT_ENDPOINT = '/dataset/commit'
 QUERY_ENDPOINT = '/dataset/query'
-CREATE_ENDPOINT = '/dataset/create'
```
TheFRedFox, line 19: Is there any plan to change now `/dataset/new_object` to `/dataset/create` again? Could be a nicer name and endpoint but would for sure bring some more work in.

TheFRedFox (approved):
Looks good.

adminapi/dataset/__init__.py
kofrezo: There is currently none but I like the idea. But lets wait sometime to do so to avoid confusion and avoid weird mistakes where people use the old endpoint with new code.

TheFRedFox: Definitely. Would go for another deprecation phase after some time. ;)

PR #113, opened 2020-10-06
Remove deprecated query function

kofrezo:


TheFRedFox (approved):
All checks went through, seems there is no other using of `query` internally.

PR #114, opened 2020-10-08
Increase hostname length to 254 characters

kofrezo:
- RFC1123: Software SHOULD handle hostnames up to 255 characters
- RFC1035: The maximum length of a DNS name is 255 octets

We allow 254 characters because one character is for the root zone
(last dot e.g. www.microsoft.com.) in DNS names and we only allow
characters via HOSTNAME_VALIDATORS that use 1 byte in UTF-8


TheFRedFox (approved):
Python-wise it looks good.

raimund-s (approved):
Setting the limit 254 makes sense from my point of view.

PR #116, opened 2020-10-13
Fix missing special attributes lead to server error

kofrezo:
Hostname and servertype are mandatory special attributes that must be
set when creating a new object.
    
We check for the attribute names to be present in the submitted request
but not if they have a value. intern_ip is optional by now as we support
servertype without intern_ip.

### DIFF
--- a/serveradmin/serverdb/query_committer.py
+++ b/serveradmin/serverdb/query_committer.py
@@ -206,17 +206,15 @@ def _delete_servers(changed, deleted, deleted_servers):
 def _create_servers(attribute_lookup, created):
     created_servers = {}
     for attributes in created:
-        if 'hostname' not in attributes:
+        if not attributes.get('hostname'):
             raise CommitError('"hostname" attribute is required.')
         hostname = attributes['hostname']
 
-        if 'servertype' not in attributes:
+        if not attributes.get('servertype'):
             raise CommitError('"servertype" attribute is required.')
         servertype = _get_servertype(attributes)
 
-        if 'intern_ip' not in attributes:
-            raise CommitError('"intern_ip" attribute is required.')
-        intern_ip = attributes['intern_ip']
+        intern_ip = attributes.get('intern_ip')
 
         attributes = dict(_get_real_attributes(attributes, attribute_lookup))
         _validate_real_attributes(servertype, attributes)


serveradmin/serverdb/query_committer.py
```diff
@@ -206,17 +206,17 @@ def _delete_servers(changed, deleted, deleted_servers):
 def _create_servers(attribute_lookup, created):
     created_servers = {}
     for attributes in created:
-        if 'hostname' not in attributes:
+        if 'hostname' not in attributes or not attributes['hostname']:
```
meikomeik, line 209: Could be written as:
`if not attributes.get('hostname'):`

serveradmin/serverdb/query_committer.py
```diff
@@ -206,17 +206,17 @@ def _delete_servers(changed, deleted, deleted_servers):
...
             raise CommitError('"hostname" attribute is required.')
         hostname = attributes['hostname']
 
-        if 'servertype' not in attributes:
+        if 'servertype' not in attributes or not attributes['servertype']:
```
meikomeik, line 213: same as above:
`if not attributes.get('servertype'):`

serveradmin/serverdb/query_committer.py
TheFRedFox, line 209: This would also be a possible solution and might be easier to read:
```suggestion
        if not attributes.get('hostname'):
```

serveradmin/serverdb/query_committer.py
TheFRedFox, line 213: Same here:
```suggestion
        if not attributes.get('servertype'):
```

serveradmin/serverdb/query_committer.py
```diff
@@ -206,17 +206,17 @@ def _delete_servers(changed, deleted, deleted_servers):
...
             raise CommitError('"servertype" attribute is required.')
         servertype = _get_servertype(attributes)
 
-        if 'intern_ip' not in attributes:
-            raise CommitError('"intern_ip" attribute is required.')
-        intern_ip = attributes['intern_ip']
+        intern_ip = None
+        if 'intern_ip' in attributes:
+            intern_ip = attributes['intern_ip']
```
TheFRedFox, line 219: Could be replaced with:
```suggestion
        intern_ip = attributes.get('intern_ip')
```
The `dict.get` method is giving the value for the key or `None` if the key does not exist.

meikomeik (changes requested):
left comments

serveradmin/serverdb/query_committer.py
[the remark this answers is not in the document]
meikomeik: one additional thing. Maybe we should also check if `intern_ip` is a parseable ip address. If it is not we could also raise an error early.

meikomeik (approved):
I like it

TheFRedFox (approved):

PR #117, opened 2020-10-13
Fix supernet attribute breaks optional intern_ip

kofrezo:
- Fix supernet attribute requires intern_ip
- Deny to make supernet attribute mandatory if intern_ip is optional (Admin)
- Deny to make intern_ip optional if any supernet attribute is mandatory (Admin)

### DIFF
--- a/serveradmin/serverdb/forms.py
+++ b/serveradmin/serverdb/forms.py
@@ -1,0 +1,43 @@
+from django import forms
+from django.core.exceptions import ValidationError
+from django.utils.translation import gettext as _
+
+
+class ServertypeAdminForm(forms.ModelForm):
+    def clean(self):
+        if self.cleaned_data.get('ip_addr_type') == 'null':
+            supernet_and_required = self.instance.attributes.filter(
+                required=True,
+                attribute__type='supernet'
+            ).only('attriute_id').values_list('attribute_id', flat=True)
+            if supernet_and_required.exists():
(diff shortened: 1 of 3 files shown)


serveradmin/serverdb/forms.py
```diff
@@ -0,0 +1,42 @@
...
+                    params={
+                        'attrs': ','.join(supernet_and_required)
+                    })
+        super().clean()
+
+
+class ServertypeAttributeAdminForm(forms.ModelForm):
+    def clean(self):
+        if (
+            self.instance.attribute.type == 'supernet' and
```
TheFRedFox, line 29: I would suggest to introduce another variable with a meaningful name just like above. So, it is easier to read the purpose of the condition statement.

kofrezo: fair enough.

serveradmin/serverdb/forms.py
```diff
@@ -0,0 +1,43 @@
...
+                    code='invalid',
+                    params={
+                        'attrs': ','.join(supernet_and_required)
+                    })
+        super().clean()
+
+
+class ServertypeAttributeAdminForm(forms.ModelForm):
+    def clean(self):
+        supernet_and_requried = (
```
TheFRedFox, line 28:
```suggestion
        supernet_and_required = (
```

TheFRedFox (approved):

PR #118, opened 2020-10-21
Fix ServertypeAttribute has no attribute

kofrezo:
If the attribute is just added with the current request instance is not
present yet so we have to use cleaned_data.


lacunoc (approved):

TheFRedFox (approved):
Oh yeah, good one. I should have seen that at the very first PR.

PR #119, opened 2020-10-26
Fix server errors (HTTP 500) due to duplicate intern_ip attribute values

kofrezo:
When cloning an object with the Servershell do not clone the intern_ip.
It must always be uniq and will conflict on save we are better of by
forcing the user to choose a IP address in the first place.

### DIFF
--- a/serveradmin/servershell/views.py
+++ b/serveradmin/servershell/views.py
@@ -379,12 +379,16 @@ def new_object(request):
 @login_required
 def clone_object(request):
     try:
+        cloned_attributes = list(Attribute.specials)
+        # intern_ip is usually unique (except for loadbalancers) therefore it
+        # makes sense to not clone it.
+        cloned_attributes.remove('intern_ip')
+        cloned_attributes.extend(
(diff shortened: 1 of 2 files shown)


serveradmin/servershell/views.py
```diff
@@ -379,12 +379,15 @@ def new_object(request):
 @login_required
 def clone_object(request):
     try:
+        cloned_attributes = list(Attribute.specials)
+        # intern_ip must be uniq - this would always conflict on save
```
meikomeik, line 383: I think this comment is misleading. intern_ip must not be unique, we have a lot of loadbalancers which share a intern_ip with another loadbalancer. I still does make sense to empty this attribute during cloning.

kofrezo: Good catch! I found that out later and already adjusted the commit message but forgot about the comment in the code.

meikomeik (approved):

TheFRedFox (approved):
Looks good.

PR #120, opened 2020-11-03
Provide API endpoint to lock something for n secs

kofrezo:


meikomeik (approved):
Looks beautiful. Approved

serveradmin/api/api.py
```diff
@@ -0,0 +1,37 @@
...
+    :param identifier: A unique identifier (e.g. 10.0.0.1)
+    :param seconds: seconds until the lock expires
+
+    :return: True on success or seconds left if already in use
+    """
+
+    # Use hash sum because this has a constant length
+    hash_sum = hashlib.sha1(str(identifier).encode()).hexdigest()
+
+    # Remove expired logs first
```
TheFRedFox, line 26:
```suggestion
    # Remove expired locks first
```

serveradmin/api/api.py
```diff
@@ -0,0 +1,37 @@
...
+    Lock.objects.filter(until__lt=timezone.now()).delete()
+
+    in_use = Lock.objects.filter(hashsum=hash_sum).only('until')
+    if in_use.exists():
+        return (in_use.get().until - timezone.now()).seconds
+
+    until = timezone.now() + timedelta(seconds=seconds)
+    Lock.objects.create(hashsum=hash_sum, duration=seconds, until=until)
```
TheFRedFox, line 34: This should most likely be in a locking transaction, so the reading of the existence and the actual getting resp. the creation of a new one is protected for Race Condition.

TheFRedFox: The `select_for_update` should then be used, too.
https://docs.djangoproject.com/en/3.1/ref/models/querysets/#select-for-update

serveradmin/api/api.py
```diff
@@ -0,0 +1,37 @@
...
+    distributed clients can check whatever something is in use or not.
+
+    :param identifier: A unique identifier (e.g. 10.0.0.1)
+    :param seconds: seconds until the lock expires
+
+    :return: True on success or seconds left if already in use
+    """
+
+    # Use hash sum because this has a constant length
+    hash_sum = hashlib.sha1(str(identifier).encode()).hexdigest()
```
TheFRedFox, line 24: This should be part of the `Lock` model class. It's in the scope of responsibility of the model.

TheFRedFox, line 24: I guess the likeliness of getting the same hash for two different identifiers is quite unlikely?

serveradmin/api/api.py
TheFRedFox: Whereby, the Race Condition is not a Security risk here actually as follow up queries would result into db errors. Still might be interesting for UX purposes.

serveradmin/api/api.py
```diff
@@ -0,0 +1,37 @@
...
+
+    # Remove expired logs first
+    Lock.objects.filter(until__lt=timezone.now()).delete()
```
TheFRedFox, line 27: I might guess that the locking can be crucial in timing, therefore it might be rather interesting to put that into a command and calling it with a cron job here and then to slim down the view function.

TheFRedFox (changes requested):
Discussions about Race Condition, Scope of Responsibility and Slimming down the function.

serveradmin/api/api.py
kofrezo: Yes, I think this is good enough.

serveradmin/api/api.py
kofrezo: I thought about this too in the first place but decided against it. 

We don't use the API yet executing this query takes at most a few milli seconds (for a few hundred rows). I think implementing a management command to run it as a cron job would be overkill here and more error prone (people can forget to setup the cron) and this can be done easily if it turns out to be a problem.

serveradmin/api/api.py
kofrezo: Why would this be the responsibility of the model it already expects a hash_sum and therefore the name of the attribute no ?

serveradmin/api/api.py
kofrezo: True, might be worth it here.

kofrezo: `select_for_update` would not have been worked because the objects did not exist before. However I found a way to make the creation of the object atomic, this should do the trick.

serveradmin/api/api.py
kofrezo: Clarified in a personal call. I added a static method to the model that can be used to avoid multiple implementations.

serveradmin/api/api.py
```diff
@@ -0,0 +1,35 @@
...
+
+    :return: True on success or seconds left if already in use
+    """
+
+    # Remove expired locks first
+    Lock.objects.filter(until__lt=timezone.now()).delete()
+
+    # Use hash sum because this has a constant length
+    hash_sum = Lock.get_hash_sum(identifier)
+    obj, created = Lock.objects.get_or_create(hash_sum=hash_sum)
```
TheFRedFox, line 27: Definitely a better approach! Still a Race Condition can happen where A is creating the Lock entity, then B is getting  this Lock entity here and processing further before A is saving the actual information for it.
Would then result in for B getting wrong result during the `return`.

`get_or_create` should however also support the direct filling of extra fields.
https://docs.djangoproject.com/en/3.1/ref/models/querysets/#django.db.models.query.QuerySet.get_or_create
Check the `defaults` parameter:

```suggestion
    obj, created = Lock.objects.get_or_create(hash_sum=hash_sum, defaults={'duration': seconds, 'until': timezone.now() + timedelta(seconds=seconds)})
```
With that it should be definitely safe enough.

serveradmin/api/api.py
TheFRedFox: Fair enough

serveradmin/api/models.py
```diff
@@ -0,0 +1,18 @@
...
+
+class Lock(models.Model):
+    class Meta:
+        db_table = 'api_lock'
+
+    hash_sum = models.CharField(max_length=40, null=False, unique=True)
+    until = models.DateTimeField(null=True)
+    duration = models.PositiveIntegerField(null=True)
+
+    @staticmethod
```
TheFRedFox, line 14: I would suggest to define it as a `@classmethod` so it is overridable for a super class.

serveradmin/api/migrations/0001_api_lock.py
```diff
@@ -0,0 +1,26 @@
+# Generated by Django 2.2.15 on 2020-11-02 13:15
```
TheFRedFox, line 1: Uhm, we now have two identical migrations here.

TheFRedFox (changes requested):
Some comments and one duplication.

serveradmin/api/migrations/0001_api_lock.py
kofrezo: Fxied

serveradmin/api/api.py
```diff
@@ -0,0 +1,34 @@
...
+    Lock.objects.filter(until__lt=timezone.now()).delete()
+
+    # Use hash sum because this has a constant length
+    hash_sum = Lock.get_hash_sum(identifier)
+    obj, created = Lock.objects.get_or_create(hash_sum=hash_sum, defaults={
+        'duration': seconds,
+        'until': timezone.now() + timedelta(seconds=seconds)})
+
+    if created:
+        return True
```
TheFRedFox, line 32: One more thing which might be of interest here. Not sure about the possibilities with the `@api_function`, but I would also use the response status to clarify what happened.
E.g. for the creation a response status `201` would be interesting and if the lock is already locked, there is `423` which is exactly `Locked`, which would fit perfectly here.

So, maybe you want to change the status codes too. Then the user can directly check for the status code instead of trying to interpret the content.

TheFRedFox (commented):
Looks good now. Just one more discussion and I will approve.

serveradmin/api/api.py
kofrezo: I like the general idea but unfortunately this does not make sense for the Python Adminapi as we use it right now because it is built in a way that users do not have to deal with the HTTP response itself. 

We could on the long term think about providing a general HTTP API which supports that.

TheFRedFox (approved):

PR #121, opened 2020-11-03
Add method to get exclusive free ip address

kofrezo:
Provide a method to get a IP address for a network that is locked for
60 seconds to avoid conflicting commits of new objects.

### DIFF
--- a/adminapi/dataset.py
+++ b/adminapi/dataset.py
@@ -8,10 +8,11 @@ from ipaddress import IPv4Address, IPv4Network, IPv6Address, IPv6Network
 from itertools import chain
 from types import GeneratorType
 
+from adminapi import api
 from adminapi.datatype import validate_value, json_to_datatype
 from adminapi.filters import Any, BaseFilter, ContainedOnlyBy
 from adminapi.request import send_request, json_encode_extra
-from adminapi.exceptions import DatasetError
+from adminapi.exceptions import DatasetError, AdminapiException
 
 NEW_OBJECT_ENDPOINT = '/dataset/new_object'
 COMMIT_ENDPOINT = '/dataset/commit'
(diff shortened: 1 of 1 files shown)


adminapi/dataset.py
```diff
@@ -230,6 +231,32 @@ def get_free_ip_addrs(self):
...
+
+        lock_api = api.get('api')
+        try:
+            for free_ip in self.get_free_ip_addrs():
+                success = lock_api.lock(free_ip)
+
+                if success is True:
+                    return free_ip
+        except StopIteration:
+            raise AdminapiException('No free IPs left!')
```
meikomeik, line 258: Why use the try: except: block here?
I think we can remove it:

```
        for free_ip in self.get_free_ip_addrs():
            if lock_api.lock(free_ip) is True:
                return free_ip

        raise AdminapiException('No free IPs left!')
```

meikomeik (changes requested):

adminapi/dataset.py
TheFRedFox: Meik is right there. Interesting comment on Stackoverflow on that: https://stackoverflow.com/a/14413978
`for` is checking for `StopIteration` and will just stop. So Meik's suggestion is correct.
```suggestion
        or free_ip in self.get_free_ip_addrs():
            if lock_api.lock(free_ip) is True:
                return free_ip

        raise AdminapiException('No free IPs left!')
```
If `lock_api.lock(free_ip)` is also not returning anything else truelike, then the explicit `is True` might also be not necessary:
```suggestion
        for free_ip in self.get_free_ip_addrs():
            if lock_api.lock(free_ip):
                return free_ip

        raise AdminapiException('No free IPs left!')
```

adminapi/dataset.py
```diff
@@ -230,6 +231,32 @@ def get_free_ip_addrs(self):
...
+        can set lock to false but may need to handle the CommitError with the
+        duplicate IP on your own.
+
+        :param lock: Lock free IP address for 60 seconds
+
+        :return:
+        """
+
+        if not lock:
+            return next(self.get_network_ip_addrs())
```
TheFRedFox, line 248: However, this can actually produce a `StopIteration` exception if the generator/iterator does not have any `next`. You might want to catch that here then.

TheFRedFox (changes requested):

adminapi/dataset.py
meikomeik: `lock_api.lock(free_ip)` returns a number or True so we need to check for exactly `True` here

adminapi/dataset.py
kofrezo: Indeed a more speaking exception might be helpful - will adjust.

PR #122, opened 2020-11-26
WIP: Fix submitting datetime in Servershell causes Server Error

kofrezo:
- Use same string representation in UI to avoid confusion
- Convert string values to desired data type (e.g. datetime)
- Fix submitting datetime in Servershell causes server error

### DIFF
--- a/adminapi/dataset.py
+++ b/adminapi/dataset.py
@@ -9,7 +9,7 @@ from itertools import chain
 from types import GeneratorType
 
 from adminapi import api
-from adminapi.datatype import validate_value, json_to_datatype
+from adminapi.datatype import validate_value, json_to_datatype, str_to_datatype
 from adminapi.filters import Any, BaseFilter, ContainedOnlyBy
 from adminapi.request import send_request, json_encode_extra
 from adminapi.exceptions import DatasetError, AdminapiException
@@ -417,6 +417,14 @@ class DatasetObject(dict):
                 self.old_values[key] = old_value
 
     def __setitem__(self, key, value):
+        if type(value) is str:
+            # We don't know the data type of the attribute here.
+            #
+            # If the attribute value is a string we try to match it against
+            # our known data types (True = 1, 0, true, false) etc. and convert
+            # it. If nothing matches we assume it is a string.
+            value = str_to_datatype(value)
+
         if isinstance(self[key], MultiAttr):
             value = MultiAttr(value, self, key)
         elif isinstance(value, GeneratorType):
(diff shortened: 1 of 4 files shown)


kofrezo:
This PR tackles 3 things:

1. Use the same string representation everywhere (edit, inspect and Servershell) e.g. `2020-11-26 10:15:00+0100`
2. Convert string input automatically to desired data type e.g. `host['migrated'] = '2020-11-26 10:15:00+0100'` 
3. Fix server error when saving hosts with datetime attribute in edit mask

adminapi/dataset.py
```diff
@@ -417,6 +417,14 @@ def _save_old_value(self, key):
                 self.old_values[key] = old_value
 
     def __setitem__(self, key, value):
+        if type(value) is str:
+            # We don't know the data type of the attribute here.
+            #
+            # If the attribute value is a string we try to match it against
+            # our known data types (True = 1, 0, true, false) etc. and convert
+            # it. If nothing matches we assume it is a string.
+            value = str_to_datatype(value)
```
TheFRedFox, line 426: Is this really safe without taking the type of field into account? Isn't it possible to store `1` or `true` as a real string?

TheFRedFox (commented):
2 discussion points. Rest looks good!

adminapi/dataset.py
kofrezo: In short: No it is not 100% safe but it "should" work for our setup.

In details:

We currently don't know the data type on the client side. If we would want to know the data type we would have to Query it at some point from the backend which can be quite expensive. I decided against this because it is a big change with huge impact.

I can happen for example that somebody creates a string attribute and does not properly set the regex validation and tries to store "false" in it.

I want to refactor this again later by either finding a cheap way to know the data type or find a another solution but this is out of the scope of this PR. Right now we can't commit  datetimes in the Servershell edit page at all.

serveradmin/servershell/templatetags/servershell.py
```diff
@@ -0,0 +1,49 @@
...
+    :param field_type:
+    :return:
+    """
+
+    if field_type == 'datetime':
+        if value.tzinfo is None:
+            value.replace(tzinfo=timezone.utc)
+        return value.astimezone(timezone.utc).strftime('%Y-%m-%d %H:%M:%S%z')
+    elif field_type == 'boolean':
+        return str(value).lower()
```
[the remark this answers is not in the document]
kofrezo: Good idea.

adminapi/dataset.py
TheFRedFox: Ok, got it. Then good to go for now.

kofrezo:
The automatic type conversion has problems in the Servershell views for editing. However the string representation still has value so I made a dedicated PR for it and will come up with a better solution in another PR.

-> https://github.com/innogames/serveradmin/pull/123

PR #124, opened 2020-12-04
Add decorator to log function execution time

kofrezo:
We already log the execution of API endpoints but don't know how much
time is spent on blocking signals. This function allows us to do so in
our custom apps.

### DIFF
--- a/serveradmin/common/utils.py
+++ b/serveradmin/common/utils.py
@@ -1,10 +1,13 @@
 """Serveradmin
 
-Copyright (c) 2019 InnoGames GmbH
+Copyright (c) 2020 InnoGames GmbH
 """
-
-import re
+import functools
 import os
+import re
+import time
+import logging
+
 from base64 import b64encode
 
 _hostname_re = re.compile(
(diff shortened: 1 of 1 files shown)


serveradmin/common/utils.py
```diff
@@ -28,3 +31,18 @@ def random_alnum_string(length):
...
+
+def profile(func):
+    @functools.wraps(func)
+    def wrapper_profile(*args, **kwargs):
+        start = time.perf_counter()
+        value = func(*args, **kwargs)
+        end = time.perf_counter()
+
+        duration = end - start
+        msg = 'Signal %s processed in %s seconds' % (func.__name__, duration)
```
TheFRedFox: Get rid of the old format style. ;)

```suggestion
        msg = 'Signal {} processed in {} seconds'.format(func.__name__, duration)
```

If we are over 3.5 the new f format is even possible:
```suggestion
        msg = f'Signal {func.__name__} processed in {duration} seconds'
```

TheFRedFox: I find the wording `Signal` weird here. Something like `Function processing` or the like would make more sense to me.

serveradmin/common/utils.py
```diff
@@ -28,3 +31,18 @@ def random_alnum_string(length):
...
+def profile(func):
+    @functools.wraps(func)
+    def wrapper_profile(*args, **kwargs):
+        start = time.perf_counter()
+        value = func(*args, **kwargs)
+        end = time.perf_counter()
+
+        duration = end - start
+        msg = f'{func.__name__} processed in {duration} seconds'
+        logging.getLogger('profile').info(msg)
```
TheFRedFox, line 45: I guess it would be interesting to make the log level customisable. The style of parametrized decorators is unfortunately not that clear. If you think it would already be cool to have it, then you might want to investigate further for parametrized decorator .. or later.

TheFRedFox (approved):

PR #125, opened 2020-12-21
Show app and owner in Servershell history

kofrezo:
One often wants to know which app (token) has made changes additionally
to the owner of the app since one owner may have multiple apps.


Woellchen (approved):
thank you very much :)

PR #126, opened 2021-01-14
Do not crash if the key is in wrong format

Felixoid:
With yubikey PIV two keys are provided:

```
> ssh-add -l
256 SHA256:ZohTkUG5ko5xzL0/ZcLX5t8m8ybEiXoCOdLeKVERxsE Public key for PIV Authentication (ECDSA)
2048 SHA256:a2L69IagYYYfJ1PDYhrj0GLT0i3EZa3mleARfdMaTZc Public key for PIV Attestation (RSA)
```

And the first one is unable to sign the message for some reason. But it's not the reason to crush.

### DIFF
--- a/adminapi/request.py
+++ b/adminapi/request.py
@@ -75,6 +75,16 @@ def calc_message(timestamp, data=None):
     return str(timestamp) + (':' + data) if data else str(timestamp)
 
 
+def signable(key):
+    """Checks if key is able to sign the message
+    """
+    try:
+        key.sign_ssh_data(b'')
+        return True
+    except SSHException:
+        return False
+
+
 def calc_signature(private_key, timestamp, data=None):
     """Create a proof that we posess the private key
 
@@ -158,7 +168,7 @@ def _build_request(endpoint, get_params, post_params):
     else:
         try:
             agent = Agent()
-            agent_keys = agent.get_keys()
+            agent_keys = filter(signable, agent.get_keys())
         except SSHException:
             raise AuthenticationError('No token and ssh agent found')
 
--- a/setup.py
+++ b/setup.py
@@ -59,6 +59,10 @@ if __name__ == '__main__':
                 'adminapi=adminapi.__main__:main',
             ],
         },
+        install_require=[
+            'paramiko',
+            'netaddr',
+        ],
         author='InnoGames System Administration',
         author_email='it@innogames.com',
     )


adminapi/request.py
```diff
@@ -91,7 +91,14 @@ def calc_signature(private_key, timestamp, data=None):
     Returns the signature as base64 encoded unicode, ready for transport.
     """
     message = calc_message(timestamp, data)
-    sig = private_key.sign_ssh_data(message.encode())
+    try:
+        sig = private_key.sign_ssh_data(message.encode())
+    except SSHException as e:
+        # The key is not necessary could be used for authentication
```
axxelG, line 97: That sentence does not make sense. What do you want to say?
"The key provided cannot be used for authentication"?

Felixoid: As PGP keys, they can be used for signature, encryption, or authentication. That's why the fact that the key is found in the agent doesn't mean, it must be able to sign something.

adminapi/request.py
```diff
@@ -91,7 +91,14 @@ def calc_signature(private_key, timestamp, data=None):
     Returns the signature as base64 encoded unicode, ready for transport.
     """
     message = calc_message(timestamp, data)
-    sig = private_key.sign_ssh_data(message.encode())
+    try:
+        sig = private_key.sign_ssh_data(message.encode())
+    except SSHException as e:
+        # There is a possbility, that key is unable to be used for auth
```
axxelG, line 97: Typo: possbility -> possibility
Better English: "There is a possibility that this key cannot be used for auth"

Felixoid: Thank you

adminapi/request.py
```diff
@@ -91,7 +91,14 @@ def calc_signature(private_key, timestamp, data=None):
     Returns the signature as base64 encoded unicode, ready for transport.
     """
     message = calc_message(timestamp, data)
-    sig = private_key.sign_ssh_data(message.encode())
+    try:
+        sig = private_key.sign_ssh_data(message.encode())
+    except SSHException as e:
+        # There is a possbility, that the key cannot be used for authentication
+        # e.g., RSA keys for PIV Attestation
```
kofrezo, line 98: In the PR conversation you mention that you fail to sign the request with the ECDSA key but in this comment you are mentioning the RSA key. With which key does it fail - maybe we can trace the root cause ?

adminapi/request.py
```diff
@@ -91,7 +91,14 @@ def calc_signature(private_key, timestamp, data=None):
...
+        if e.args[0] == 'key cannot be used for signing':
+            return ""
```
kofrezo, line 100: I am not sure if returning an empty string here is a good idea. I am not even sure if this can be processed by the backend or if it would reject it. Can you please check ?

kofrezo (changes requested):

kofrezo:
As discussed in person I first want to make sure that the keys not able to sign are "valid" and not a behaviour/bug in the integration of the YubiKey (e.g. YubiKey loading too many/wrong keys like e.g. for signing instead of authentication). 

If it is a possible use case to have SSH keys loaded in the ssh-agent which are not able to sign data we should try to filter them out in the first place if possible instead of letting them fail in the earliest spot possible.

https://github.com/innogames/serveradmin/blob/master/adminapi/request.py#L161

setup.py
```diff
@@ -59,6 +59,12 @@
                 'adminapi=adminapi.__main__:main',
             ],
         },
+        extras_require={
```
kofrezo, line 62: Good catch to add these. But shouldn't this be done via `install_requires` since adminapi does not work without them and `extas_requires` lists optional requirements ?

> A dictionary mapping names of “extras” **(optional features of your project)** to strings or lists of strings specifying what other distributions must be installed to support those features. See the section on Declaring required dependency for details and examples of the format of this argument.

Source: https://setuptools.readthedocs.io/en/latest/references/keywords.html (extras_require)

adminapi/request.py
```diff
@@ -75,6 +75,17 @@ def calc_message(timestamp, data=None):
     return str(timestamp) + (':' + data) if data else str(timestamp)
 
 
+def signable(key):
+    """
```
kofrezo, line 79: Please make the docstring start here.

kofrezo (changes requested):

setup.py
Felixoid: The confusion here is different packages at one setup. The requirements are for `adminapi`, but here is `serveradmin` itself too. So, I don't know how to handle it propely

adminapi/request.py
Felixoid: :+1:

setup.py
kofrezo: Ah yes. I remember the problem - true.

However in this case I think it is fine because in the backend (Serveradmin) we need both paramiko and netaddr:
- https://github.com/innogames/serveradmin/blob/master/serveradmin/apps/models.py#L10
- https://github.com/innogames/serveradmin/blob/master/serveradmin/serverdb/models.py#L12

So I suggest we put it to `install_requires`. Does that make sense to you ?

Felixoid: Sounds perfect :+1:

kofrezo (approved):

PR #128, opened 2021-02-08
Fix Servershell attr command not working

### LINKED ISSUE #127: Servershell attr command not working
I wanted to give Serveradmin a try this weekend.  The new interface looks neat.  Good job!

I added a few attributes, servertypes, and objects.  Object creating went smooth.  Then I realized the "attr" command not working.  Is it a known issue?

I looked at my JavaScript console, but haven't seen anything other than "Failed to load resource: the server responded with a status of 404 (Not Found) http://127.0.0.1:8000/static/js/popper.min.js.map".

This is Safari Version 14.0.2 (16610.3.7.1.9) on MacOS 11.1 here.

kofrezo:
Having 2 javascript files with the same name (command.js) seems to be a
problem for the Safari Browser on Mac OS. It seems to load the wrong
file which breaks the auto completion.

fixes #127


TheFRedFox:
If this is really fixing the problem, then fine. ^^

TheFRedFox (approved):

kofrezo:
This will not solve the problem. See the issue for details.

PR #130, opened 2021-02-16
Allow inspect by hostname

kofrezo:
The `/inspect` endpoint is only used by the Servershell inspect command with the object_id. 

However some users might find it useful to be able to setup a custom search engine in their browser and be able to use the inspect endpoint to display details of a Server by the hostname.

### DIFF
--- a/serveradmin/servershell/views.py
+++ b/serveradmin/servershell/views.py
@@ -1,6 +1,6 @@
 """Serveradmin - Servershell
 
-Copyright (c) 2020 InnoGames GmbH
+Copyright (c) 2021 InnoGames GmbH
 """
 
 import json
@@ -10,10 +10,18 @@ from itertools import islice
 from django.contrib import messages
 from django.contrib.auth.decorators import login_required
 from django.core.exceptions import (
-    ObjectDoesNotExist, PermissionDenied, ValidationError
+    ObjectDoesNotExist,
+    PermissionDenied,
+    ValidationError
 )
-from django.http import HttpResponse, HttpResponseRedirect, Http404, \
-    JsonResponse, HttpResponseBadRequest, HttpResponseNotFound
+from django.http import (
+    HttpResponse,
+    HttpResponseRedirect,
+    Http404,
+    JsonResponse,
+    HttpResponseBadRequest,
+    HttpResponseNotFound
+)
 from django.shortcuts import redirect
 from django.template.response import TemplateResponse
 from django.urls import reverse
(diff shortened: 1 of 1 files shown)


serveradmin/servershell/views.py
```diff
@@ -197,8 +198,21 @@ def get_results(request):
 
 @login_required
 def inspect(request):
-    server = Query({'object_id': request.GET['object_id']}, None).get()
-    return _edit(request, server, template='inspect')
+    if request.method != 'GET':
```
TheFRedFox, line 201: There are the http decorators, which you can use for that: https://docs.djangoproject.com/en/2.2/topics/http/decorators/

serveradmin/servershell/views.py
```diff
@@ -197,8 +198,21 @@ def get_results(request):
...
+        return HttpResponseNotAllowed(['GET'])
+
+    if 'hostname' in request.GET:
```
TheFRedFox, line 204: If `hostname` AND `object_id` are specified in the GET request, wouldn't it make more sense to favor the `object_id` over the `hostname` as it would be more precise?

serveradmin/servershell/views.py
```diff
@@ -13,7 +13,8 @@
     ObjectDoesNotExist, PermissionDenied, ValidationError
 )
 from django.http import HttpResponse, HttpResponseRedirect, Http404, \
-    JsonResponse
+    JsonResponse, HttpResponseBadRequest, HttpResponseNotAllowed, \
```
TheFRedFox, line 16: Normally, we are using braces in favor of backslashes:
```python
from django.http import (
    HttpResponse, ....
    Foobar
)
```
or the like.

serveradmin/servershell/views.py
kofrezo: Nice - didn't not stumble across them yet.

serveradmin/servershell/views.py
kofrezo: As you can see it was like this already and my IDE just sticks to whatever it was before.

serveradmin/servershell/views.py
kofrezo: I will swap them since that makes more sense but as pointed out in the description of the PR this endpoint is really just used by the Servershell which always uses the object_id (pro grammatically) and this addition for the hostname is just a convenience feature for some power users who really should know what they are doing.

On the very long term I aim for getting rid of the speciality of the hostname attribute anyways. We already store plenty of information about projects, networks etc. where we "abuse" the hostname as an identifier. So this will probably go away or change in the future is should just be seen as a intermediate code for power users.

serveradmin/servershell/views.py
kofrezo: I will make a dedicated commit to clean this mess up.

TheFRedFox: Sure, no need to do it in this PR.

serveradmin/servershell/views.py
TheFRedFox: Regards of the long term it makes even more sense to switch it, so it won't be a behavior change when you fill in both query parameters.

TheFRedFox (approved):

Felixoid (approved):

PR #131, opened 2021-02-16
Implement testing with github actions

Felixoid:
Travis works worse and worse with time. This PR implements testing in github actions


Felixoid:
Does it look like Travis stopped working already? :laughing: 

They give only 1k minutes monthly for open source projects..

kofrezo (approved):
I checked the output of the new tests. Looks good - many thanks.

PR #132, opened 2021-02-18
Several quick performance improvements

Woellchen:
This PR heavily improves the performance of some aspects of Serveradmin with minimal effort.
1. Improves query times for queries involving relations and for nested queries.
2. Improves loading time of object history.

Query times were improved by ~40-45% and history loading times were improved by more than 90%.
While I still see lots of potential for further optimizations, I think with these minimal changes the gain is pretty massive.
Disclaimer: The improvements were calculated from my local instance :) I hope they will also be noticable as much on production.
See commit messages for all details.

### DIFF
--- a/serveradmin/serverdb/query_executer.py
+++ b/serveradmin/serverdb/query_executer.py
@@ -258,6 +258,6 @@ def _get_servers(filters, attribute_lookup, related_vias):
     # easy: get and execute the raw SQL query.
     sql_query = get_server_query(attribute_filters, related_vias)
     try:
-        return list(Server.objects.raw(sql_query))
+        return list(Server.objects.defer('intern_ip').raw(sql_query))
     except DataError as error:
         raise ValidationError(error)
--- a/serveradmin/serverdb/query_materializer.py
+++ b/serveradmin/serverdb/query_materializer.py
@@ -136,7 +136,10 @@ class QueryMaterializer:
                 for sa in ServerRelationAttribute.objects.filter(
                     value_id__in=self._server_attributes.keys(),
                     attribute_id__in=reversed_attributes.keys(),
-                ).select_related('server'):
+                ).select_related('server').defer(
+                    'server__intern_ip',
+                    'server__servertype',
+                ):
                     self._add_attribute_value(
                         sa.value,
                         reversed_attributes[sa.attribute_id],
@@ -147,7 +150,10 @@ class QueryMaterializer:
                 for sa in ServerAttribute.get_model(key).objects.filter(
                     server__in=self._server_attributes.keys(),
                     attribute__in=attributes,
-                ).select_related('server'):
+                ).select_related('server').defer(
+                    'server__intern_ip',
+                    'server__servertype',
+                ):
                     self._add_attribute_value(
                         sa.server,
                         attribute_lookup[sa.attribute_id],
@@ -221,7 +227,10 @@ class QueryMaterializer:
         for sa in ServerAttribute.get_model(attribute.type).objects.filter(
             server__hostname__in=servers_by_related.keys(),
             attribute=attribute,
-        ).select_related('server'):
+        ).select_related('server').defer(
+            'server__intern_ip',
+            'server__servertype',
+        ):
             for target in servers_by_related[sa.server]:
                 self._add_attribute_value(target, attribute, sa.get_value())
 
--- a/serveradmin/serverdb/query_materializer.py
+++ b/serveradmin/serverdb/query_materializer.py
@@ -100,7 +100,7 @@ class QueryMaterializer:
             sa = ServertypeAttribute.objects.filter(
                 servertype_id=sa.servertype_id,
                 attribute_id=related_via_attribute_id,
-            ).select_related('attribute').get()
+            ).prefetch_related('attribute').get()
             self._select_servertype_attribute(sa.attribute, sa)
 
     def _initialize_attributes(self, servers_by_type):
@@ -136,7 +136,7 @@ class QueryMaterializer:
                 for sa in ServerRelationAttribute.objects.filter(
                     value_id__in=self._server_attributes.keys(),
                     attribute_id__in=reversed_attributes.keys(),
-                ).select_related('server').defer(
+                ).prefetch_related('server').defer(
                     'server__intern_ip',
                     'server__servertype',
                 ):
@@ -150,7 +150,7 @@ class QueryMaterializer:
                 for sa in ServerAttribute.get_model(key).objects.filter(
                     server__in=self._server_attributes.keys(),
                     attribute__in=attributes,
-                ).select_related('server').defer(
+                ).prefetch_related('server').defer(
                     'server__intern_ip',
                     'server__servertype',
                 ):
@@ -227,7 +227,7 @@ class QueryMaterializer:
         for sa in ServerAttribute.get_model(attribute.type).objects.filter(
             server__hostname__in=servers_by_related.keys(),
             attribute=attribute,
-        ).select_related('server').defer(
+        ).prefetch_related('server').defer(
             'server__intern_ip',
             'server__servertype',
         ):
--- a/serveradmin/serverdb/views.py
+++ b/serveradmin/serverdb/views.py
@@ -8,9 +8,10 @@ import dateparser
 from django.conf import settings
 from django.contrib import messages
 from django.contrib.auth.decorators import login_required
+from django.contrib.auth.models import AbstractUser
 from django.core.exceptions import ObjectDoesNotExist
 from django.core.paginator import Paginator, EmptyPage, PageNotAnInteger
-from django.db.models import Q
+from django.db.models import Prefetch, Q
 from django.http import Http404
 from django.shortcuts import get_object_or_404, redirect
 from django.template.response import TemplateResponse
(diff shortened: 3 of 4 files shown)


serveradmin/serverdb/views.py
```diff
@@ -87,38 +100,35 @@ def history(request):
...
-    adds = adds.select_related()
-    updates = updates.select_related()
-
     if commit_id:
         adds = adds.filter(commit__pk=commit_id)
         updates = updates.filter(commit__pk=commit_id)
         deletes = deletes.filter(commit__pk=commit_id)
 
+    related = ('commit__app', 'commit__user')
+    adds = adds.select_related(*related)
```
TheFRedFox, line 123: I am not familiar with the django models here. Maybe it won't change anything, but before there was the `select_related()` call, which was actually doing something. If it is still necessary, I guess it should be changed to:
``` python
adds = adds.select_related().select_related(*related)
```

TheFRedFox, line 123: On another note, did you also try to use `prefetch_related` instead of `select_related`? It is using more queries, however one can also select `many_to_one` and `many_to_many` relations with that. Maybe interesting to know for future improvements on other columns.

serveradmin/serverdb/views.py
```diff
@@ -8,9 +8,10 @@
 from django.conf import settings
 from django.contrib import messages
 from django.contrib.auth.decorators import login_required
+from django.contrib.auth.models import AbstractUser
 from django.core.exceptions import ObjectDoesNotExist
 from django.core.paginator import Paginator, EmptyPage, PageNotAnInteger
-from django.db.models import Q
+from django.db.models import Prefetch, Q
```
TheFRedFox, line 14: I can't see `Prefetch` being used here? Am I blind?

serveradmin/serverdb/views.py
```diff
@@ -8,9 +8,10 @@
 from django.conf import settings
 from django.contrib import messages
 from django.contrib.auth.decorators import login_required
+from django.contrib.auth.models import AbstractUser
```
TheFRedFox, line 11: I can't see `AbstractUser` being used here? Am I blind?

serveradmin/serverdb/views.py
```diff
@@ -87,38 +100,35 @@ def history(request):
...
+
     change_list = ([('add', obj) for obj in adds] +
                    [('update', obj) for obj in updates] +
                    [('delete', obj) for obj in deletes])
     change_list.sort(key=lambda entry: entry[1].commit.change_on, reverse=True)
 
-    for change in change_list:
-        commit = change[1].commit
-        if commit.user:
-            commit.commit_by = commit.user.get_full_name()
```
TheFRedFox, line 116: Is `commit_by` not needed anymore?

TheFRedFox (commented):
Changes look reasonable. On a general note I wonder if we can hardcode those columns there. I am not familiar with all the attributes, but I thought the attributes are all very dynamic and can be completely different in another instance.

If it is like that, we shouldn't put them hardcoded there but rather get those things from settings.

Additionally, one error in the commit description (which were btw very awesome!), a drop from 360s to 32s is not "~90% faster" but more like "~1000% faster"/"10 times faster" the 90% are correct in a wording like: "which reduces the time by ~90%".

serveradmin/serverdb/views.py
Woellchen: the whole loop wasn't needed anymore because the template directly accesses the objects, which call `__str__`. this must be a relic

serveradmin/serverdb/views.py
Woellchen: you are generally right. but this is a one-to-one relation and we can save an additional query here compared to `prefetch_related`. i still highly recommend `prefetch_related` for all many relations (see `serveradmin/serverdb/query_materializer.py`)

Woellchen: i don't think this would work. `select_related` returns related objects. in this case `commit`/`ChangeCommit`. but i also want to "prefetch" commit's relations, too, to save on many unnecessary queries. because the template will try to access `app` and `user` of `commit` and this will trigger twice as many separate queries as there are commit rows. by passing those related fields explicitly, you can optimize it to query everything at once

TheFRedFox: I understand that you want to add the related fields from commit, too, and that it is needed. My comment was not removing that. My comment was stating, that by removing the `select_related()` call, we have a behavior change as it won't select other relations automatically anymore. From django:
```
In these cases it is possible to call select_related() with no arguments. This will follow all non-null foreign keys it can find - nullable foreign keys must be specified.
```
It is anyway not recommended and might not be necessary anyway (as I said, I am not familiar with the models here), just want to point out that it might be a behavior change and if we still need it, it should be:
``` python
adds = adds.select_related().select_related(*related)
```
If we don't need the other relations, which would be selected by `select_related()`, then ignore that comment. ^^

TheFRedFox: I understand, that for one-to-one relations one should use `select_related`, I was just curious how much slower `prefetch_related` would be compared to `select_related`. ;) Was asking if just by any chance you tested that. If not, then not. ^^

Woellchen: i got what you meant. what i mean is that chaining those calls does not have any effect :D if you first select all related and afterwards only a subset of it, then it will choose the subset only anyway

Woellchen: yeah i tested it in a way that i saw an extra unnecessary query. i haven't measured it though. it probably finishes very fast in any case

TheFRedFox: After call it is now clear. `select_related()` is not creating a list but just setting a property to `True`, which will be overwritten by the next explicit `select_related(...)` call. The documentation is a bit misleading there, as it sounds like that the `select_related()` call will create a list directly and not just set a property to `True`.
So, the chaining would not make sense.

TheFRedFox (approved):
All discussion points were solved. 👍🏼 
Again, kudos for the descriptions. :)

HerrSpace (commented):
The changes changes (hehe) are pretty straight forward. Love it.

---

This use of prefetch_related is interesting. Previously my mental model was:
- use select_related for FKs
- use prefetch_related for many to many

Most (all?) of the things you prefetch_related are FKs though.

Thinking about this change a bit, I see however that using prefetch_related for the EA part of our EAV database table is probably beneficial, as we only query server and attribute once (per typed attribute table, per recursion level, but still). We also only create a single python object for those rows and hence only de-serialize expensive attributes like intern_ip and relations once. I've not actually tested this, but it makes sense in my head 😬🙃

---

Frankly I think you should consider rewriting serveradmin without an EAV model in the medium term. It sounds crazy complex, but really the only crazy complex part of serveradmin is this EAV spaghetti code. The question then becomes whether you want to keep serveradmins flexibility of changing the schema without requiring migrations, but that's probably best discussed elsewhere.

HerrSpace:
Man. This beast was fun to work on.

serveradmin/serverdb/query_materializer.py
```diff
@@ -136,7 +136,10 @@ def _add_attributes(self, servers_by_type):
                 for sa in ServerRelationAttribute.objects.filter(
                     value_id__in=self._server_attributes.keys(),
                     attribute_id__in=reversed_attributes.keys(),
-                ).select_related('server'):
+                ).prefetch_related('server').defer(
+                    'server__intern_ip',
+                    'server__servertype',
+                ):
```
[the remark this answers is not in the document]
HerrSpace: To make it more clear, I mean something like this (syntax and schema might be wrong, I haven't touched this in 1.5 years):

```python
from adminapi.dataset import Query

list(Query(
    {
        'servertype': 'vm',
    }, [
        'hostname',
        'intern_ip',
        {'loadbalancer': [
            'hostname',
            'intern_ip'
        ]}
    ]
))
```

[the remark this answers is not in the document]
Woellchen: i tested your query and the time difference is not really measurable. but what happens indeed is that two more queries are fired now. one for retrieving all ips of all related objects and one for retrieving servertypes for the nested query.
at least django is smart enough to batch all servers together to retrieve a deferred column.
so while this adds unnecessary overhead i would like to not optimize it any further in this pr because i see the root problem in the design of the query logic.
fixing this properly imo requires rewriting the query classes in a way that they know what they are actually querying upfront, which will lead to more improvements where we could shrink down the total amount of queries made in general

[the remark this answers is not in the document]
HerrSpace: Cool! I agree as long as the deferred attribute queries get batched it doesn't really matter. :)

serveradmin/serverdb/query_executer.py
```diff
@@ -258,6 +258,6 @@ def _get_servers(filters, attribute_lookup, related_vias):
     # easy: get and execute the raw SQL query.
     sql_query = get_server_query(attribute_filters, related_vias)
     try:
-        return list(Server.objects.raw(sql_query))
+        return list(Server.objects.defer('intern_ip').raw(sql_query))
```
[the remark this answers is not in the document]
Woellchen: actually in this case it only defers the ip object creation which saves a lot of time because django is not parsing them. this is because the intern_ip is already inside `sql_query` and will always be selected. see my comment below for why i do not want to touch this query logic now :)

Woellchen:
thank you very much @HerrSpace!

serveradmin/serverdb/query_executer.py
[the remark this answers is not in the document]
HerrSpace: Ack!

kofrezo:
> On a general note I wonder if we can hardcode those columns there. I am not familiar with all the attributes, but I thought the attributes are all very dynamic and can be completely different in another instance.

If you are referring to hard coding `intern_ip` and `servertype_id` attributes: This is fine for now.

We are actually working on getting rid of `intern_ip` as special field on the server column but there is more to do in advance. 

For `servertype_id` this is just fine this one is never going to go away.

serveradmin/serverdb/views.py
```diff
@@ -87,38 +100,35 @@ def history(request):
...
 
-    adds = adds.select_related()
-    updates = updates.select_related()
-
     if commit_id:
         adds = adds.filter(commit__pk=commit_id)
         updates = updates.filter(commit__pk=commit_id)
         deletes = deletes.filter(commit__pk=commit_id)
 
+    related = ('commit__app', 'commit__user')
```
kofrezo, line 122: Ahh I wasn't aware that prefetching relations does not cascade to other relations - makes perfect sense.

Good catch! :+1:

kofrezo (approved):
Awesome findings!

Took me quite some time to understand why the history loading part was slow before. Makes perfect sense to specifically specify all the relations to prefetch and that this does not happen implicit.

Also for the `prefetch_related` vs `select_related` this makes perfect sense with the growing amount of object with relations we have. I executed it on an old dataset and there actually the difference was almost zero but on the current dataset when fetching many objects with many relations its a huge improvement I measuered even 50% performance gain.

PR #133, opened 2021-02-23
Fix autocomplete in safari webkit

kofrezo:


serveradmin/servershell/static/js/servershell/command.js
```diff
@@ -614,6 +614,16 @@ $(document).ready(function() {
...
 
+        // Workaround:
+        //
+        // The input change event is not fired or somehow interrupted in
+        // Safari, Epihany and other like browsers, they all have in common to
+        // have "AppleWebKit/605.1.15" in the user agent.
+        //
+        // We just trigger the event manually here to make it work for those.
+        // This will fire it twice for other browser but that does not harm.
+        $('#command').change();
```
TheFRedFox, line 625: Just as a side note, not mandatory for the PR as I don't see a doubled event being a problem here: 
With `navigator.userAgent` you can get the User Agent of the browser with JS. One should rather use feature detection e.g. with [Modernizr](https://modernizr.com), however in that case it actually might not really be possible.

TheFRedFox (approved):
Just a little side note.

serveradmin/servershell/static/js/servershell/command.js
kofrezo: I tried that. Unfortunately the user agent does not differ. I can't however say if it would be possible with modernizr.

PR #135, opened 2021-03-09
Fix auto completion overrides (similar) previous attrs

kofrezo:
When building the results for the auto completion we take the full
entered search term and replace the entered term to complete with the
autocompleted part. Because we replace the first occurence this can break
if the term already contains similar words like for example
"game_market=de game_<autocomplete>".

This commit fixes it by easily search for the last occurence assuming
people only use auto completion at the end and don't jump to the
beginning of the search and enter something and expect auto completion.


lamaral:
I have lost count of how many times the `st` in `stretch` got replaced when the autocompletion for `state` hit :joy:

TheFRedFox (approved):
Looks correct.

PR #137, opened 2021-03-09
Remove wrong timer for spinner

kofrezo:


Felixoid:
So, it will be just removed?

Would it make sense to keep only the final time summary?

Felixoid:
IMHO, the problem with the spinner is inaccurate result time. At least for me.

I don't insist, just asking, BTW

TheFRedFox (approved):

PR #138, opened 2021-03-09
Show current version in Servershell

kofrezo:
Show current version in the top right of the navigation.


serveradmin/common/templatetags/common.py
```diff
@@ -1,13 +1,16 @@
...
 """
 
 from django import template
 
+from adminapi import VERSION
+
 register = template.Library()
 
 
+
```
TheFRedFox, line 13: This looks like unnecessary line

PR #139, opened 2021-03-10
Add simple spinner timer

kofrezo:
Show only total time spend on an action.

### DIFF
--- a/serveradmin/common/static/js/serveradmin.js
+++ b/serveradmin/common/static/js/serveradmin.js
@@ -9,40 +9,30 @@ $(document).ready(function() {
     // everywhere when ever we need it for example when doing long running
     // ajax requests for the servershell search.
     window.spinner = {
-        _running: 0,
+        _startedAt: 0,
         _spinner: $('#spinner'),
-        _toggle_css: function() {
-            if (this._spinner.css('display') === 'none') {
-                this._spinner.show();
-            }
-
-            if (this._spinner.hasClass('text-success')) {
-                this._spinner.removeClass('text-success');
-                this._spinner.addClass('text-secondary');
-
-                this._spinner.css('animation-play-state', 'paused');
-            }
-            else {
-                this._spinner.removeClass('text-secondary');
-                this._spinner.addClass('text-success');
-
-                this._spinner.css('animation-play-state', 'running');
-            }
-        },
         enable: function () {
-            this._running++;
-            if (this._running > 1) {
-                return;
-            }
+            console.debug('spinner enabled');
+            this._startedAt = Date.now();
 
-            this._toggle_css();
+            this._spinner.removeClass('text-secondary');
+            this._spinner.addClass('text-success');
+            this._spinner.css('animation-play-state', 'running');
         },
         disable: function () {
-            this._running--;
+            console.debug('spinner disabled');
 
-            if (this._running === 0) {
-                this._toggle_css();
+            let elapsed = Date.now() - this._startedAt;
+            if (elapsed > 1000) {
+                $('#spinner-timer-value').html(`${elapsed / 1000} s`);
             }
+            else {
+                $('#spinner-timer-value').html(`${elapsed} ms`);
+            }
+
+            this._spinner.css('animation-play-state', 'paused');
+            this._spinner.removeClass('text-success');
+            this._spinner.addClass('text-secondary');
         },
 
     }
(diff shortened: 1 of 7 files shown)


serveradmin/common/static/js/serveradmin.js
```diff
@@ -11,38 +11,28 @@ $(document).ready(function() {
...
         },
         disable: function () {
-            this._running--;
+            console.debug('spinner disabled');
 
-            if (this._running === 0) {
-                this._toggle_css();
+            let elapsed = Date.now() - this._running;
+            if (elapsed > 1000) {
+                $('#spinner-timer-value').html(`${elapsed / 1000} s`);
```
TheFRedFox, line 27: Just the seconds or seconds and milliseconds? Would it be ok from UX perspective if it would show `1` for `1.999` seconds?

serveradmin/common/static/js/serveradmin.js
```diff
@@ -11,38 +11,28 @@ $(document).ready(function() {
...
-                this._spinner.css('animation-play-state', 'running');
-            }
-        },
         enable: function () {
-            this._running++;
-            if (this._running > 1) {
-                return;
-            }
+            console.debug('spinner enabled');
+            this._running = Date.now();
```
TheFRedFox, line 16: As `running` sounds more like a boolean value I suggest to use something more like `startedAt` or the like if `running` is not used somewhere else already.

serveradmin/servershell/static/js/servershell/autocomplete/search.js
```diff
@@ -50,8 +50,6 @@ $(document).ready(function () {
         delay: 250, // Wait n ms before starting to auto complete to avoid needles requests to backend
         autoFocus: true,
         source: function (request, response) {
-            spinner.enable();
```
TheFRedFox, line 53: Comment for the whole commit:
Looks like it would make sense in future to implement something more global to use, something like:
``` js
Serveradmin.sendCommand(...) // sendRequest // request // or the like
```
Which would then do the calls to the spinner, so you don't need to put it to every single space.

serveradmin/common/static/js/serveradmin.js
```diff
@@ -11,38 +11,28 @@ $(document).ready(function() {
...
 
-            this._toggle_css();
+            this._spinner.removeClass('text-secondary');
+            this._spinner.addClass('text-success');
+            this._spinner.css('animation-play-state', 'running');
         },
         disable: function () {
-            this._running--;
```
TheFRedFox, line 55: To avoid calling functions more often, I suggest to still use `running` or some other boolean value to check if the spinner is actually running and can/need to be disabled.
Same could work for `enable` function.

TheFRedFox (commented):
Just a couple of discussion points before approval.

serveradmin/common/static/js/serveradmin.js
kofrezo: Fair point. I will change it.

serveradmin/servershell/static/js/servershell/autocomplete/search.js
kofrezo: I thought about this but in some places we do $.ajax and in some $.getJSON and settings vary greatly so if there is not method to hook into the XHR requests I don't think it is worth it.

serveradmin/common/static/js/serveradmin.js
kofrezo: I would like to keep it simple here. It should be in the responsibility of the developer to start and stop only one action at a time. Everything else can be a good hint to a bug (e.g. spinner keeps running).

serveradmin/common/static/js/serveradmin.js
kofrezo: Adjusted.

serveradmin/servershell/static/js/servershell/autocomplete/search.js
TheFRedFox: With a appropriate service class/object we could still achieve it. But that's something for another time. ;)

serveradmin/common/static/js/serveradmin.js
```diff
@@ -9,40 +9,30 @@ $(document).ready(function() {
     // everywhere when ever we need it for example when doing long running
     // ajax requests for the servershell search.
     window.spinner = {
-        _running: 0,
+        _startedAt: 0,
```
TheFRedFox, line 12: The appropriate initial value for a date would most probably then be `null`.

TheFRedFox (approved):
One minor thing. Early approve.

serveradmin/common/static/js/serveradmin.js
kofrezo: In Javascript it doesn't make any difference:
```
> let now = Date.now()
> now - 0
< 1615449781944
> now - null 
< 1615449781944
```

PR #140, opened 2021-03-12
Make auto completion delay configurable per user

kofrezo:
Allow to configure delay before auto completion kicks in per user and
save configuration in session.


TheFRedFox (approved):
Looks good.

PR #141, opened 2021-03-12
Concatenate and compress JS/CSS on the fly

kofrezo:
Use Django compressor app to create a single compressed JS and CSS file.

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "3b9a095a8daf8cd367167e218141eed621d3d51e8b7a0a945bed40eeeb678874"
+            "sha256": "bd67e2fd5ac288bd1f19da30dc2d92736e488a8e555adf85fe2f06f1b7adecfd"
         },
         "pipfile-spec": 6,
         "requires": {
@@ -102,6 +102,21 @@
             ],
             "index": "pypi",
             "version": "==2.2.19"
+        },
+        "django-appconf": {
+            "hashes": [
+                "sha256:1b1d0e1069c843ebe8ae5aa48ec52403b1440402b320c3e3a206a0907e97bb06",
+                "sha256:be58deb54a43d77d2e1621fe59f787681376d3cd0b8bd8e4758ef6c3a6453380"
+            ],
+            "version": "==1.0.4"
+        },
+        "django-compressor": {
+            "hashes": [
+                "sha256:57ac0a696d061e5fc6fbc55381d2050f353b973fb97eee5593f39247bc0f30af",
+                "sha256:d2ed1c6137ddaac5536233ec0a819e14009553fee0a869bea65d03e5285ba74f"
+            ],
+            "index": "pypi",
+            "version": "==2.4"
         },
         "django-netfields": {
             "hashes": [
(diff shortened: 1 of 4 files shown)


kofrezo:
I noticed some errors in the Javascript where whitespaces are stripped/compressed from at least interpolated strings. I will fix that first:

```
# Original
choices.push({
    'label': `Attr: ${attribute_id}`,
    'value': `${command} ${attribute_id}`,
});               

# Compressed
choices.push({
    'label': `Attr:${attribute_id}`,
    'value': `${command}${values_string}${attribute_id}`,
})
```

Pipfile.lock
```diff
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "325a3b81f83f06bc671c6af64fe576536ad4a8eea219dee603136d58feab5a78"
```
TheFRedFox, line 4: I am not sure right now, if it is possible with pipenv, but if, it would be better to just have the changes of the install in here without any other updates.

serveradmin/common/templates/base.html
```diff
@@ -76,5 +80,6 @@
 <script src="{{ STATIC_URL }}js/serveradmin.js"></script>
 <script src="{{ STATIC_URL }}js/plugins/terminal_keyboard.js"></script>
 {% block additional_scripts %}{% endblock %}
```
TheFRedFox, line 82: How much sense does it make to also include any additional stuff to the caching too? If we don't have that many additional stuff it might be ok.

serveradmin/settings.py
```diff
@@ -90,6 +91,13 @@
 # Examples: "http://media.lawrence.com/media/", "http://example.com/media/"
 MEDIA_URL = '/media/'
 
+# Add compressor to static file finders to allow compressing CSS/JS on the fly
```
TheFRedFox, line 94: Have you also considered to use the offline compressing?

TheFRedFox:
> I noticed some errors in the Javascript where whitespaces are stripped/compressed from at least interpolated strings. I will fix that first:
> 
> ```
> # Original
> choices.push({
>     'label': `Attr: ${attribute_id}`,
>     'value': `${command} ${attribute_id}`,
> });               
> 
> # Compressed
> choices.push({
>     'label': `Attr:${attribute_id}`,
>     'value': `${command}${values_string}${attribute_id}`,
> })
> ```

You might want to try another compressor then: https://django-compressor.readthedocs.io/en/latest/settings/
The default one seems to use http://opensource.perlig.de/rjsmin/

Didn't check it, but maybe e.g. the YUI one is doing it better.

serveradmin/common/templates/base.html
kofrezo: We have quite some for example in the Servershell and Resources page and we never know what will be added:

- https://github.com/innogames/serveradmin/blob/master/serveradmin/servershell/templates/servershell/index.html#L7
- https://github.com/innogames/serveradmin/blob/master/serveradmin/resources/templates/resources/index.html#L8

Also I think there is value in including these because it guarentess that browser download the latest version if something has changed in there.

serveradmin/settings.py
kofrezo: Yes I discarded it because of the possible Caveats. Especially the first bullet point sounded dangerous to me:

- https://django-compressor.readthedocs.io/en/stable/scenarios/#id1

Pipfile.lock
kofrezo: I tried and failed. It seems to automatically update all dependecies that are not pinned to a specific version as soon as I run `pipenv update` or `pipenv lock`

serveradmin/settings.py
kofrezo: Same as above. Quite a lot in my opinion and if it were only for making sure that users get the latest version when something changes (Browser Cache).

- https://github.com/innogames/serveradmin/blob/master/serveradmin/servershell/templates/servershell/index.html#L170

serveradmin/common/templates/base.html
kofrezo: Same as above. Quite a lot in my opinion and if it were only for making sure that users get the latest version when something changes (Browser Cache).

- https://github.com/innogames/serveradmin/blob/master/serveradmin/servershell/templates/servershell/index.html#L170

kofrezo:
> > I noticed some errors in the Javascript where whitespaces are stripped/compressed from at least interpolated strings. I will fix that first:
> > ```
> > # Original
> > choices.push({
> >     'label': `Attr: ${attribute_id}`,
> >     'value': `${command} ${attribute_id}`,
> > });               
> > 
> > # Compressed
> > choices.push({
> >     'label': `Attr:${attribute_id}`,
> >     'value': `${command}${values_string}${attribute_id}`,
> > })
> > ```
> 
> You might want to try another compressor then: https://django-compressor.readthedocs.io/en/latest/settings/
> The default one seems to use http://opensource.perlig.de/rjsmin/
> 
> Didn't check it, but maybe e.g. the YUI one is doing it better.

A good idea. I will try this first before adjusting the code.

kofrezo:
This is acutally fixed in the latest 2.4.x version of django-compressor. Debian just ships a very old one 2.2.x ... I will try to build an up-to-date one.

Pipfile.lock
TheFRedFox: Pity, it behaves like that. Alright  then.

serveradmin/common/templates/base.html
TheFRedFox: Uhh, there is even dynamic javascript in there. This should result in quite some amount of cache files. I would suggest to monitor that if that could result in a problem. Good for me to try out though.

serveradmin/settings.py
TheFRedFox: Yeah, I saw that some javascript has dynamic values, so javascript will get generated during runtime. Due to CSP and general Good Practice we anyway should try and avoid such javascript generation. So, offline compression might come back later then.

TheFRedFox:
> This is acutally fixed in the latest 2.4.x version (which requires rjsmin 1.1.0) of django-compressor. Debian just ships a very old one 2.2.x ... I will try to build an up-to-date one.

Should I see the more restrict dependency version definition in the Pipfile as the solution for this PR? Fine for me then.

TheFRedFox (approved):
Looks good, just the comment again to monitor the amount of created compressed files as we have some dynamic generated javascript code.

kofrezo:
> Looks good, just the comment again to monitor the amount of created compressed files as we have some dynamic generated javascript code.

Yes you do it is already in there: https://github.com/innogames/serveradmin/pull/141/files#diff-230078d672f10d17463a8a6265cad825b790885898256a3365be90685caac58dR28

serveradmin/common/templates/base.html
kofrezo: Why would this be a problem ? 

If you think in terms of size: We have maybe 2 handful of pages and the Javascript and CSS files are way less than a 1 MB. Every log file we have is bigger in size.

serveradmin/settings.py
kofrezo: > Due to CSP and general Good Practice
I lack to imagine a concrete for an attack vector for dynamic Javscript generation. Do you have some reference for this ?

PR #144, opened 2021-03-24
Safely add attribute values to prevent XSS

kofrezo:


serveradmin/servershell/static/js/servershell/result.js
```diff
@@ -84,24 +84,40 @@ get_row_html = function(object, number) {
                     let to_delete = change.remove.join(', ');
                     let value = object[attribute_id].filter(v => !to_delete.includes(v)).join(', ');
 
-                    cell.html(`${value} <del>${to_delete}</del> <u>${to_add}</u>`)
+                    cell.text(value);
+
+                    let del = $('<del></del>');
```
TheFRedFox, line 89: Btw, for creating just an empty html elements you don't need the closing tag:
``` js
$('<del>')
```
is sufficient already.

serveradmin/servershell/static/js/servershell/result.js
```diff
@@ -84,24 +84,40 @@ get_row_html = function(object, number) {
...
+                    cell.append(del);
+
+                    let add = $('<u></u>');
+                    add.text(to_add);
+                    cell.append(add);
                 }
                 else {
                     let to_delete = change.old === null ? '' : change.old;
                     let new_value = change.new === undefined ? '': change.new;
-                    cell.html(`<del>${to_delete}</del>&nbsp;<u>${new_value}</u>`);
```
TheFRedFox, line 89: I'm not sure, but maybe you will use that `&nbsp;` whitespace. Is that ok?

TheFRedFox (approved):
Just 2 minor discussions. Nothing for rejection. Already approve.

serveradmin/servershell/static/js/servershell/result.js
kofrezo: Ah neat. I will adjust the code.

serveradmin/servershell/static/js/servershell/result.js
kofrezo: When including the `&nbsp` the jQuery methods fail so I decided to remove it. We should do it with CSS anyways if it bothers people but I think it looks ok.

PR #147, opened 2021-04-07
Replace (possibly) XSS vunerable with safe Javscript

kofrezo:
Follow up to https://github.com/innogames/serveradmin/pull/144

@TheFRedFox found some more spots where we can possibly run into XSS vunerabilities. This PR mitigates them by safely adding the content as text instead of html or removes the code completely.

### DIFF
--- a/serveradmin/servershell/static/js/servershell.js
+++ b/serveradmin/servershell/static/js/servershell.js
@@ -216,7 +216,7 @@ servershell.alert = function(text, level, auto_dismiss=5) {
     box.addClass(`alert-${level}`);
 
     let message = $('#js-alert-message');
-    message.html(text);
+    message.text(text);
 
     box.toggle();
 
--- a/serveradmin/servershell/static/js/servershell.js
+++ b/serveradmin/servershell/static/js/servershell.js
@@ -192,7 +192,7 @@ servershell.pages = function() {
  * don't have the backend. If the error is coming from the backend please
  * using Django messages mechanism. It will generate bootstrap alert as well.
  *
- * @param text message to show (can be html)
+ * @param text message to show
  * @param level either primary,secondary,success,danger,warning,info,light,dark
  * @param auto_dismiss seconds to wait before disappear (0 never)
  */
--- a/serveradmin/servershell/static/js/servershell/result.js
+++ b/serveradmin/servershell/static/js/servershell/result.js
@@ -15,7 +15,7 @@ servershell.update_result = function() {
     header.empty();
     header.append('<th scope="col"></th>');
     header.append('<th scope="col">#</th>');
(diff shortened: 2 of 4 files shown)


serveradmin/servershell/static/js/servershell.js
```diff
@@ -216,7 +216,7 @@ servershell.alert = function(text, level, auto_dismiss=5) {
     box.addClass(`alert-${level}`);
 
     let message = $('#js-alert-message');
-    message.html(text);
+    message.text(text);
```
TheFRedFox, line 219: Looking at line 195 here, the documentation says that it can have html. I wonder if it was somewhere already used with html.
My suggestion would be that it could still have html, but not by default and one need to add a parameter like `html=true` to tell the method to use `html` and not `text`.

serveradmin/servershell/static/js/servershell/result.js
```diff
@@ -59,8 +59,8 @@ get_row_html = function(object, number) {
     }
 
     // Standard columns which should always be present
-    row.append(`<td><input tabindex="3" type="checkbox" name="server" value="${object.object_id}"/></td>`);
-    row.append(`<td>${number + servershell.offset}</td>`);
+    row.append($('<td>').append($('<input tabindex="3" type="checkbox" name="server"/>').val(object.object_id)));
```
TheFRedFox, line 62: If that is possible like that, then it looks awesome! 👍🏼 
I'm just not sure if in the end just the `input` gets appended to the row and not the `td`. Did you test it already on staging?

TheFRedFox (commented):
Awesome changes! Just have 2 discussion points to resolve first.

serveradmin/servershell/static/js/servershell.js
kofrezo: Only place that I can see that **could** return HTML would be:

- https://github.com/innogames/serveradmin/blob/master/serveradmin/servershell/static/js/servershell/command.js#L585

I am in favour for security here and better just allow only plain text. I will adjust the doc instead.

serveradmin/servershell/static/js/servershell/result.js
kofrezo: I double checked it - it works.

Screenshot from 2021-04-08 10-31-13

TheFRedFox: Awesome! 👍🏼

serveradmin/servershell/static/js/servershell.js
TheFRedFox: If everything just uses `text` then even better. Sure text is just safer. Just saying, if html is needed, I would go for the opt-in html parameter. Adjusting the doc for now is totally good.

TheFRedFox (approved):

PR #149, opened 2021-04-21
Extend validation for inet attributes

kofrezo:
For historical reasons we had validation only for the special attribute intern_ip but no (conistent) validation for intern_ip and inet attribute itself and together (combination). This PR aims to bring a consistent validation behaviour for all inet attributes and the special attribute intern_ip based on the ip_addr_type configuration of a servertype.

It includes tests and some comment on how we can further cleanup this when we refactor the validation and get rid of the special intern_ip later.

### DIFF
--- a/serveradmin/serverdb/fixtures/auth_user.json
+++ b/serveradmin/serverdb/fixtures/auth_user.json
@@ -1,0 +1,20 @@
+[
+    {
+          "model": "auth.user",
+          "pk": 1,
+          "fields": {
+              "password": "pbkdf2_sha256$150000$w82dDYYVQ5kk$tZWXPj0KeKhtxjKTINpZdaVBjvJit31vUvKnsc2KS3s=",
+              "last_login": "2021-04-16T09:25:40.336Z",
+              "is_superuser": true,
+              "username": "admin",
+              "first_name": "",
+              "last_name": "",
+              "email": "",
+              "is_staff": true,
+              "is_active": true,
+              "date_joined": "2020-11-05T12:42:44.952Z",
+              "groups": [],
+              "user_permissions": []
+          }
+    }
+]
--- a/serveradmin/serverdb/forms.py
+++ b/serveradmin/serverdb/forms.py
@@ -1,55 +1,35 @@
 from django import forms
 from django.core.exceptions import ValidationError
-from django.utils.translation import gettext as _
+
+from serveradmin.serverdb.models import ServertypeAttribute
 
 
 class ServertypeAdminForm(forms.ModelForm):
-    def clean(self):
-        if self.cleaned_data.get('ip_addr_type') == 'null':
-            supernet_and_required = self.instance.attributes.filter(
-                required=True,
-                attribute__type='supernet'
-            ).only('attriute_id').values_list('attribute_id', flat=True)
-            if supernet_and_required.exists():
-                raise ValidationError(_(
-                        'ip_addr_type for Servertype must not be "null" '
-                        'if any attribute of type supernet is required '
-                        'but %(attrs)s is/are!'
-                    ),
-                    code='invalid',
-                    params={
-                        'attrs': ','.join(supernet_and_required)
-                    })
-        super().clean()
+    pass
 
 
 class ServertypeAttributeAdminForm(forms.ModelForm):
-    def clean(self):
-        supernet_and_required = (
-            self.cleaned_data['attribute'].type == 'supernet' and
-            self.cleaned_data['servertype'].ip_addr_type == 'null' and
-            self.cleaned_data.get('required') is True
-        )
-        if supernet_and_required:
-            raise ValidationError(_(
-                    'Attributes of type %(type)s can not be required if '
-                    'ip_addr_type of Servertype is null!'
-                ),
-                code='invalid',
-                params={
-                    'attribute': self.cleaned_data['attribute'].attribute_id,
-                    'type': self.cleaned_data['attribute'].type,
-                })
+    class Meta:
+        model = ServertypeAttribute
+        fields = [
+            'attribute',
+            'related_via_attribute',
+            'consistent_via_attribute',
+            'required',
+            'default_value',
+            'default_visible',
+        ]
 
-        # It makes no sense to add inet attributes to hosts of ip_addr_type
-        # null because they would have to be empty anyways.
+    def clean(self):
+        # It makes no sense to add inet or supernet attributes to hosts of
+        # ip_addr_type null because they would have to be empty anyways.
         inet_attribute = (
-                self.cleaned_data['attribute'].type == 'inet' and
-                self.cleaned_data['servertype'].ip_addr_type == 'null'
+            self.cleaned_data['attribute'].type in ('inet', 'supernet') and
+            self.instance.servertype.ip_addr_type == 'null'
         )
         if inet_attribute:
-            raise ValidationError(_(
-                'Adding an attribute of type inet with ip_addr_type: null is'
-                ' not possible!'), code='invalid')
+            raise ValidationError(
+                'Adding an attribute of type inet or supernet when '
+                'ip_addr_type is null is not possible!')
 
         super().clean()
--- a/serveradmin/serverdb/forms.py
+++ b/serveradmin/serverdb/forms.py
@@ -40,4 +40,16 @@ class ServertypeAttributeAdminForm(forms.ModelForm):
                     'attribute': self.cleaned_data['attribute'].attribute_id,
                     'type': self.cleaned_data['attribute'].type,
                 })
+
+        # It makes no sense to add inet attributes to hosts of ip_addr_type
+        # null because they would have to be empty anyways.
+        inet_attribute = (
+                self.cleaned_data['attribute'].type == 'inet' and
+                self.cleaned_data['servertype'].ip_addr_type == 'null'
+        )
+        if inet_attribute:
+            raise ValidationError(_(
+                'Adding an attribute of type inet with ip_addr_type: null is'
+                ' not possible!'), code='invalid')
+
         super().clean()
--- a/serveradmin/serverdb/models.py
+++ b/serveradmin/serverdb/models.py
@@ -323,8 +323,9 @@ class Server(models.Model):
             intern_ip__net_contains_or_equals=self.intern_ip,
         )
 
-    def clean(self, *args, **kwargs):
-        super(Server, self).clean(*args, **kwargs)
+    def clean(self):
+        super(Server, self).clean()
+
         if self.servertype.ip_addr_type == 'null':
             if self.intern_ip is not None:
                 raise ValidationError('IP address must be null.')
--- a/serveradmin/serverdb/models.py
+++ b/serveradmin/serverdb/models.py
@@ -14,7 +14,7 @@ from netaddr import EUI
 from django.db import models
 from django.core.exceptions import ValidationError
 from django.core.validators import RegexValidator
-from django.utils.timezone import now, utc
+from django.utils.timezone import now
 from django.contrib.auth.models import User
 
 import netfields
--- a/serveradmin/serverdb/models.py
+++ b/serveradmin/serverdb/models.py
@@ -38,10 +38,10 @@ ATTRIBUTE_TYPES = {
 }
 
 IP_ADDR_TYPES = [
-    ('null', 'null (intern_ip must be empty and inet attributes not be used)'),
-    ('host', 'host (intern_ip must be a host /32,/128 and unique)'),
-    ('loadbalancer', 'loadbalancer (intern_ip must be a host /32,/128)'),
-    ('network', 'network (intern_ip must be a network and not overlap)'),
+    ('null', 'null: intern_ip must be empty, no inet attributes'),
+    ('host', 'host: intern_ip and inet must be an ip address and unique across all objects'),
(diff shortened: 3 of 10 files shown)


serveradmin/serverdb/models.py
```diff
@@ -573,6 +668,28 @@ class Meta:
...
+                _('%(attribute_id)s must be null'), code='invalid value',
+                params={'attribute_id': self.attribute_id})
+        elif ip_addr_type == 'host':
+            is_ip_address(self.value)
+            is_unique(self.value)
+        elif ip_addr_type == 'loadbalancer':
+            is_ip_address(self.value)
+        elif ip_addr_type == 'network':
+            is_network(self.value)
+            network_overlaps(self.value, self.server.servertype_id)
```
TheFRedFox, line 676: We now have here more or less duplicated code. Just be aware of that. ;)

serveradmin/serverdb/models.py
```diff
@@ -38,10 +38,10 @@
...
 
 IP_ADDR_TYPES = [
-    ('null', 'null (intern_ip must be empty and inet attributes not be used)'),
-    ('host', 'host (intern_ip must be a host /32,/128 and unique)'),
-    ('loadbalancer', 'loadbalancer (intern_ip must be a host /32,/128)'),
-    ('network', 'network (intern_ip must be a network and not overlap)'),
+    ('null', 'null: intern_ip must be empty, no inet attributes'),
+    ('host', 'host: intern_ip & inet must be an ip address and unique across all objects'),
+    ('loadbalancer', 'loadbalancer: intern_ip & inet must be an ip address'),
+    ('network', 'network: intern_ip & inet must be an ip network, not overlapping with same servertype'),
```
TheFRedFox, line 45: If this is written English, then you might rather want to use `and` instead of `&`.

serveradmin/serverdb/forms.py
```diff
@@ -1,55 +1,35 @@
 from django import forms
 from django.core.exceptions import ValidationError
-from django.utils.translation import gettext as _
+
+from serveradmin.serverdb.models import ServertypeAttribute
 
 
 class ServertypeAdminForm(forms.ModelForm):
-    def clean(self):
```
TheFRedFox, line 7: Is this now safe with the new validators inside the models?

serveradmin/serverdb/models.py
```diff
@@ -426,6 +442,9 @@ def clean(self, *args, **kwargs):
                 raise ValidationError(
                     _('intern_ip must not be null'), code='missing value')
 
+            if type(self.intern_ip) not in [IPv4Interface, IPv6Interface]:
```
TheFRedFox, line 445: Why not having this check inside `inet_to_python` already? It would makes sense to check it in the function too already.

serveradmin/serverdb/tests/test_ip_addr_type.py
```diff
@@ -0,0 +1,450 @@
...
+        server = self._get_server('network')
+        server['intern_ip'] = '10.0.0.0/30'
+        server['ip_config'] = '10.0.1.0/30'
+        server.commit(user=User.objects.first())
+
+        duplicate = self._get_server('network')
+        duplicate['intern_ip'] = '10.0.2.0/30'
+        duplicate['ip_config_new'] = '10.0.1.0/30'
+        with self.assertRaises(ValidationError):
+            duplicate.commit(user=User.objects.first())
```
TheFRedFox, line 450: No new line

serveradmin/serverdb/tests/test_ip_addr_type.py
```diff
@@ -0,0 +1,450 @@
+import logging
```
TheFRedFox, line 1: While checking for something like [dataProviders](https://phpunit.readthedocs.io/en/9.5/writing-tests-for-phpunit.html#writing-tests-for-phpunit-data-providers-examples-datatest-php) in django/python I found this:
https://www.hacksoft.io/blog/improve-your-tests-django-fakes-and-factories .. maybe you want to read through the factories, could be a possibility to reduce the test code here. But nothing for the PR.

TheFRedFox (approved):
Some minor things which would be interesting to check and fix, giving already an early approve.

serveradmin/serverdb/models.py
```diff
@@ -29,7 +30,7 @@
     'relation': str,
     'reverse': str,
     'number': lambda x: float(x) if '.' in str(x) else int(x),
-    'inet': lambda x: ip_network(x) if '/' in str(x) else ip_address(x),
+    'inet': lambda x: inet_to_python(x),
```
hasegeli, line 33: Note that this was doing something else than `inet_to_python()`

`IPInterface` objects are something difficult for people to use.  They allow host addresses with network length or network addresses with host bits set, like `10.0.0.1/24`.  We allow either a host address like `10.0.0.1/32`, or a network address like `10.0.0.0/24` when `ip_addr_type` is `network`.

Postgres `inet` data type works just like `IPInterface`, so we could allow these without any extra effort, but chose to restrict things having host or network addresses explicitly to keep things more structured.

serveradmin/serverdb/tests/test_ip_addr_type.py
hasegeli: It would be nice to have a header in this file just like all others.

hasegeli: It is great to have tests now.  Maybe you can add the command to run the tests to the README.

serveradmin/serverdb/fixtures/auth_user.json
```diff
@@ -0,0 +1,20 @@
...
+              "last_name": "",
+              "email": "",
+              "is_staff": true,
+              "is_active": true,
+              "date_joined": "2020-11-05T12:42:44.952Z",
+              "groups": [],
+              "user_permissions": []
+          }
+    }
+]
```
hasegeli, line 20: No new file in here either

hasegeli (commented):
Neat work @kofrezo and review @TheFRedFox!  I was passing and could't resist the temptation to add some comments.

serveradmin/serverdb/models.py
```diff
@@ -324,7 +324,10 @@ def get_supernet(self, servertype):
         )
 
     def clean(self, *args, **kwargs):
-        super(Server, self).clean(*args, **kwargs)
+        # The super method does actually nothing but we keep this here in
+        # case this changes.
+        super(Server, self).clean()
```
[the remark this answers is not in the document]
kofrezo: >  So, I suggest to add *args and **kwargs again

Yeah this is probably safer. I removed it because the code style check made an complaint about passung unused paramters.

> Also, this comment is not very useful in here.

When thinking again about it I It was more a note/reminder to myself. I will remove it.

> If super().clean() didn't need calling, then there wouldn't be such a function.

I would like to agree but from what I can see in this particular case the function is really just present for design or interface description. From the super clean method:

```
   # django.db.models.base

    def clean(self):
        """
        Hook for doing any extra model-wide validation after clean() has been
        called on every field by self.clean_fields. Any ValidationError raised
        by this method will not be associated with a particular field; it will
        have a special-case association with the field defined by NON_FIELD_ERRORS.
        """
        pass
```

However I agree that we should call it and it makes sense to pass it args and kwargs.

serveradmin/serverdb/models.py
```diff
@@ -68,6 +69,91 @@ def get_choices(types):
...
+    """
+    prefix_length = ip_interface.network.prefixlen
+    max_prefix_length = ip_interface.network.max_prefixlen
+
+    if prefix_length != max_prefix_length:
+        raise ValidationError(
+            'Netmask length must be {0}'.format(max_prefix_length))
+
+
+def is_unique(ip_interface: Union[IPv4Interface, IPv6Interface]) -> None:
```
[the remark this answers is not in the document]
kofrezo: Yes I think this is a good improvment :+1:

serveradmin/serverdb/models.py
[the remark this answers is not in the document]
kofrezo: Looking at the Sourcecode of the base method and the documentation: https://docs.djangoproject.com/en/2.2/ref/models/instances/#django.db.models.Model.clean again I think this method should not have parameters at all. As @TheFRedFox pointed out in DM args and kwargs are not used in the methods body and I could not find any usages of it with parameters so I think it is safe to remote the parameters.

serveradmin/serverdb/models.py
```diff
@@ -68,6 +69,91 @@ def get_choices(types):
...
+    """Validate if IPv4/IPv6 address is unique
+
+    Raises a ValidationError if intern_ip or any other attribute of type inet
+    with this ip_address already exists.
+
+    :param ip_interface:
+    :return:
+    """
+
+    # We relinquish from querying the duplicate hosts here and giving the user
```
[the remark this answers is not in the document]
kofrezo: Yep I think thats better wording.

serveradmin/serverdb/models.py
```diff
@@ -68,6 +69,91 @@ def get_choices(types):
     return zip(*([types] * 2))
 
 
+# TODO: Make validators out of the methods is_ip_address, is_unique and
```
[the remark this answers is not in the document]
kofrezo: I though about this too but I use `ServerInetAttribute` and `Server` model which would introduce a circular dependency. Once could of course pass the QuerySets instead of the models directly but since I am going to refactor this again anyways I decided to make it just work for now this way.

serveradmin/serverdb/models.py
kofrezo: It is not fully duplicated but just partially. I am hoping to be able to remove the `intern_ip` attribute special code soon so that we have inet attributes only and then remove this code.

serveradmin/serverdb/models.py
kofrezo: I don't see any good reason why `and` is in preference of `&` but I also don't see any reason to discuss about this as they are effectively the same so I just adjusted this to `and`.

serveradmin/serverdb/forms.py
kofrezo: Not safe in the models but we just do not allow adding Attributes of type `inet` or `supernet` anymore in the first place therefore this cannot happen.
```
+            self.cleaned_data['attribute'].type in ('inet', 'supernet') and
+            self.instance.servertype.ip_addr_type == 'null'
```

serveradmin/serverdb/models.py
kofrezo: Because this is only the case for the special `intern_ip`. From what I could see `inet` attributes were always already transformed and I did not want to add this extra check for inet attribues. Once we remove `intern_ip` this would not be necessary anymore and this would be gone with the removal of this code block.

serveradmin/serverdb/query_committer.py
```diff
@@ -81,6 +81,25 @@ def commit_query(created=[], changed=[], deleted=[], app=None, user=None):
...
+        # API here by implementing some custom validators and building forms
+        # for the servertypes on-the-fly as described here:
+        #
+        #   - https://docs.djangoproject.com/en/2.2/ref/forms/validation/
+        #
+        # This would allow us to work with Django board tools and all it's
+        # features. Less pain because we always work around or even clash
+        # with the Django work flow and last but least allow us to use the
+        # same logic/code for the Servershell (edit, new) page and the Query
+        # engine (Web API) which currently does not use forms at all.
```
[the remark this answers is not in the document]
kofrezo: Thanks these insights are helpful. I definetely want to refactor this at a later point in time and may come back with some questions regarding the related_via and consisted_via. 

If you want you can also already elaborate this here or somewhere else so I can just read it up - however you prefer.

serveradmin/serverdb/models.py
kofrezo: Thanks for pointing this out again. I am aware that this is a change of behaviour in the method but I wanted to have one transformation method for inet and intern_ip attribute(s) and networks and hosts. The related logic for validation ensures that a host remains a host and a network a network. The overall behaviour for this should be the same if I did not miss any tests ;-)

serveradmin/serverdb/models.py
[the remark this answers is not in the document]
TheFRedFox: Fair enough. 👍🏼

serveradmin/serverdb/models.py
TheFRedFox: Then actually a little `TODO`/comment might be nice here.

serveradmin/serverdb/models.py
TheFRedFox: In "proper" written English, the `&` is very rarely used, and just used to combine nouns for a strong relationship. Got this page with a few examples: https://english.stackexchange.com/questions/3127/when-to-use-instead-of-and Therefore, I guess `and` is a better use than `&` here. But, sure, nothing to worry about for the PR.

serveradmin/serverdb/models.py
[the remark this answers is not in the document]
TheFRedFox: Here, as django is not using the abstraction library (https://docs.python.org/3/library/abc.html) I see this method being meant as an abstract method, meant to be implemented in a concrete model class.
Therefore, not calling the super-method is ok in my eyes, however to follow the inheritance and calling the super-method even though it is meant as being an abstract one, is fine, too.

As the "abstract" clean is not taking any arguments and we are not using any arguments, I would also say to actually get rid of the extra parameters in our implementation.

PR #151, opened 2021-05-11
Prevent warnings for timeouts with decimal part

axxelG:
Pylance for example generates a "reportGeneralTypeIssues" for a timout of 1.5 seconds. Adding a type hint prevents this.


axxelG:
I removed the type hint and changed the default value from `60` to `60.0`

kofrezo (approved):

PR #152, opened 2021-05-18
Fix IPv4/IPv6 comparision issues

kofrezo:
- Make sure we work with ip_interface everywhere to avoid comparision issues
- Only submit and commit changes in Servershell edit form

### DIFF
--- a/serveradmin/servershell/views.py
+++ b/serveradmin/servershell/views.py
@@ -5,6 +5,7 @@ Copyright (c) 2021 InnoGames GmbH
 
 import json
 from distutils.util import strtobool
+from ipaddress import IPv6Address, IPv4Address, ip_interface
 from itertools import islice
 
 from django.contrib import messages
@@ -273,6 +274,10 @@ def _edit(request, server, edit_mode=False, template='edit'):  # NOQA: C901
                 except ValidationError:
                     invalid_attrs.add(attribute_id)
 
+            # TODO: Remove when PR135 is merged
+            if type(current[attribute_id]) in (IPv4Address, IPv6Address):
+                current[attribute_id] = ip_interface(current[attribute_id])
+
             # Submit only changes
             if current[attribute_id] != value:
                 server[attribute_id] = value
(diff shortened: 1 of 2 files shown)


serveradmin/servershell/views.py
```diff
@@ -281,14 +281,25 @@ def _edit(request, server, edit_mode=False, template='edit'):  # NOQA: C901
             if server.object_id:
                 action = 'edited'
                 created = []
-                changed = [server._serialize_changes()]
+
+                changes = server._serialize_changes()
+                if len(changes.keys()) > 1:
+                    changed = [changes]
+                else:
+                    # Only object_id has been passed and we do not support
```
TheFRedFox, line 289: Wouldn't it be easier to strip `object_id` from the changes then?
``` python
changes.pop('object_id', None)
```
Might be more reliable for situation in which `object_id` might have not been sent.

serveradmin/servershell/views.py
```diff
@@ -281,14 +281,25 @@ def _edit(request, server, edit_mode=False, template='edit'):  # NOQA: C901
...
+                    changed = []
             else:
                 action = 'created'
                 created = [server]
                 changed = []
 
             try:
-                commit_obj = commit_query(created, changed, user=request.user)
+                if not created and not changed:
+                    messages.info(request, str('Nothing has changed.'))
```
TheFRedFox, line 299: In general, this is not needed to be inside the `try`-block.

TheFRedFox (approved):
2 minor things. Early approve.

serveradmin/servershell/views.py
kofrezo: No because this will leave me with a empty dictionary in a list: `[{}]` which will evaluate to True in the later if condition
```
>>> bool([{}])
True
```
I could check for the content in the if condition explicetly again or check for the object_id explicelty additionally to the length. We cannot change an object without sending an object_id anyways that would fail beforehand due to the failed query to get the current object.

serveradmin/servershell/views.py
kofrezo: True. Will remove it.

kofrezo:
The tests fail because this will be an API change. The client (adminapi) will no longer get IPv4/IPv6Address but IPv4/IPv6Interface. I will check if this is fine.

lamaral (approved):

PR #153, opened 2021-05-18
Specify prefix length for ip addresses of inet attributes

kofrezo:
Ensure we always work with IPv4/IPv6Interface data types for both IP addresses and networks in the backend and adminapi and return ip addresses with prefix length (e.g. `10.0.0.1/32` instead of `10.0.0.1`) for the Web API.

### DIFF
--- a/serveradmin/serverdb/query_materializer.py
+++ b/serveradmin/serverdb/query_materializer.py
@@ -16,7 +16,7 @@ from serveradmin.serverdb.models import (
     ServertypeAttribute,
     Server,
     ServerAttribute,
-    ServerRelationAttribute,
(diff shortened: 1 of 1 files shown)


kofrezo:
Note: Check why we do not already have IPv4/IPv6Interfaces. I did not find spot where we specify the prefix_length

> InetAddressField will store values in PostgreSQL as type INET. In Python, the value will be represented as an ipaddress.ip_interface object representing an IP address and netmask/prefix length pair unless the store_prefix_length argument is set to False`, in which case the value will be represented as an ipaddress.ip_address object.

-> https://pypi.org/project/django-netfields/

PR #154, opened 2021-05-19
Fix FilterValueError when creating new object

kofrezo:
With the recent changes to only submit changes values we missted that
the when creating a new object there is no current object we can compare
the values against.

### DIFF
--- a/serveradmin/servershell/views.py
+++ b/serveradmin/servershell/views.py
@@ -282,7 +282,7 @@ def _edit(request, server, edit_mode=False, template='edit'):  # NOQA: C901
                 continue
 
             if current:
-                # TODO: Remove when PR135 is merged
+                # TODO: Remove when PR153 is merged
                 if type(current[attribute_id]) in (IPv4Address, IPv6Address):
                     current[attribute_id] = ip_interface(current[attribute_id])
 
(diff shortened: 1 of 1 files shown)


serveradmin/servershell/views.py
```diff
@@ -274,12 +276,15 @@ def _edit(request, server, edit_mode=False, template='edit'):  # NOQA: C901
                 except ValidationError:
                     invalid_attrs.add(attribute_id)
 
-            # TODO: Remove when PR135 is merged
-            if type(current[attribute_id]) in (IPv4Address, IPv6Address):
-                current[attribute_id] = ip_interface(current[attribute_id])
+            if current:
```
TheFRedFox, line 279: You might want to check if you want
``` python
if attribute_id in current:
    ...
```
Just to prevent a KeyError two lines later. Possible that you want that KeyError though.

TheFRedFox (approved):
Just one thought point.

serveradmin/servershell/views.py
kofrezo: The key error could still happen for new objects when we rely on the client but I think it is still an improvement. Will add.

PR #158, opened 2021-06-02
Update Django, Jinja2 and some other libraries

kofrezo:


kofrezo:
This requires no review. Tests all succeeded and building docs with sphinx still works.

PR #159, opened 2021-06-06
Fix quotes in search term are HTML escaped

kofrezo:
The search term for the Servershell should be escaped for Javascript but
not for HTML. Otherwise when reloading the page we have an invalid
search term with HTML escape sequences.

### DIFF
--- a/serveradmin/servershell/templates/servershell/index.html
+++ b/serveradmin/servershell/templates/servershell/index.html
@@ -193,8 +193,8 @@
 
         // Initialize servershell object properties with initial values
         // from backend. This allows us setting defaults from backend.
-        servershell.term = "{{ term }}";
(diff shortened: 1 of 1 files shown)


kofrezo:
@TheFRedFox is doing some tests to make sure this does not introduce an XSS attack vector. First tests look good and even when we tried to pass URLs like:
```
?term=%2F%2F%20%3C%2Fscript%3E%3Cscript%3Econsole.log%28%27buu%27%29%3B%3C%2Fscript%3E&shown_attributes%5B%5D=hostname&shown_attributes%5B%5D=intern_ip&shown_attributes%5B%5D=servertype&shown_attributes%5B%5D=state&deep_link=true
```
The string was escaped properly.

PR #160, opened 2021-06-07
Fix additional stylesheets in wrong template block

kofrezo:
Stylesheets must go to the additional_style template block otherwise on
the-fly compression for CSS does not work.


lamaral (approved):
LGTM :)

PR #162, opened 2021-06-18
Fix inet attribute overlaps with itself

kofrezo:
When changing an inet attribute it refuses to save when the new value
overlaps with the old one because it does not exlude itself.


TheFRedFox (approved):
Makes sense.

PR #165, opened 2021-06-23
Display ip_addr_type as read-only when editing

kofrezo:
We still don't support changing the ip_addr_type of existing objects but
not being able to lookup the current type at all is not helpful so we
make it read-only in the edit form.

### DIFF
--- a/serveradmin/serverdb/admin.py
+++ b/serveradmin/serverdb/admin.py
@@ -38,7 +38,7 @@ class ServertypeAdmin(admin.ModelAdmin):
             # Because of the complexity when changing servertypes of existing
             # objects and the little use-cases we have right now we don't
             # support it.
-            fields = fields + tuple(['ip_addr_type'])
+            fields += ('ip_addr_type',)
 
         return fields
 
--- a/serveradmin/serverdb/admin.py
+++ b/serveradmin/serverdb/admin.py
@@ -31,11 +31,16 @@ class ServertypeAdmin(admin.ModelAdmin):
         ServertypeAttributeInline,
     )
 
-    def get_exclude(self, request, obj=None):
-        # Because of the great complexity when changing servertypes of existing
-        # objects and the little use-cases we have we just deny it for now.
+    def get_readonly_fields(self, request, obj=None):
+        fields = super().get_readonly_fields(request, obj)
+
         if obj:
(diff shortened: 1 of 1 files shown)


serveradmin/serverdb/admin.py
```diff
@@ -31,12 +31,15 @@ class ServertypeAdmin(admin.ModelAdmin):
...
-        # Because of the great complexity when changing servertypes of existing
-        # objects and the little use-cases we have we just deny it for now.
+    def get_readonly_fields(self, request, obj=None):
         if obj:
+            # Because of the complexity when changing servertypes of existing
+            # objects and the little use-cases we have right now we don't
+            # support it.
             return ['ip_addr_type']
 
+        return super().get_readonly_fields(request, obj)
```
TheFRedFox, line 41: You first want to get the value of the super method and then add your own on it.
``` python
def get_readonly_fields(self, request, obj=None):
    fields = super().get_readonly_fields(request, obj)
    #fields = super().get_readonly_fields(request, obj) or []
    if obj:
        fields.append('ip_addr_type')

    return fields
```

TheFRedFox (changes requested):
One detail which should be changed first.

serveradmin/serverdb/admin.py
kofrezo: Ah yes that makes sense 👍

serveradmin/serverdb/admin.py
```diff
@@ -31,11 +31,16 @@ class ServertypeAdmin(admin.ModelAdmin):
...
-        # objects and the little use-cases we have we just deny it for now.
+    def get_readonly_fields(self, request, obj=None):
+        fields = super().get_readonly_fields(request, obj)
+
         if obj:
-            return ['ip_addr_type']
+            # Because of the complexity when changing servertypes of existing
+            # objects and the little use-cases we have right now we don't
+            # support it.
+            fields = fields + tuple(['ip_addr_type'])
```
TheFRedFox, line 41:
```suggestion
            fields += ('ip_addr_type',)
```
or 
``` suggestion
            fields += tuple('ip_addr_type')
```
is possible.

TheFRedFox (approved):
One little suggestion, early approve.

PR #166, opened 2021-06-23
Set browser URL to latest Query

kofrezo:
Replace URL in browser with Query when submitting the search to avoid
loosing the selection of attributes when using multiple attributes.

### DIFF
--- a/serveradmin/servershell/static/js/servershell/search.js
+++ b/serveradmin/servershell/static/js/servershell/search.js
@@ -39,6 +39,10 @@ servershell.submit_search = function() {
             'deep_link': true,
         });
 
+        // Replace the URL with the requested one from the Ajax request to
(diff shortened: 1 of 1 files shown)


TheFRedFox (approved):
Makes sense. I just want to add the suggestion to maybe use `pushState` even. Then the queries will be navigatable and one could jump back to the previous search.

kofrezo:
Hmm not sure. I am worried we could flood the history of a user with all possible configurations just for the visibility of attributes.

kofrezo:
I adjusted it to pushState. If entries in the history turn out to much we can change it again.

kofrezo:
I tested it locally and the going back in history does not work because the Javascript history still holds the new data - I think. Anyways that was not goal of this PR so I will merge it for now.

PR #167, opened 2021-06-23
Allow bookmarking selection of attributes

kofrezo:
Add a Servershell command that either saves the current selection of
attributes visible or loads them by a given name. The selection is
persisted in the local browser storage to avoid overhead in the backend.

### DIFF
--- a/serveradmin/servershell/static/js/servershell/autocomplete/command.js
+++ b/serveradmin/servershell/static/js/servershell/autocomplete/command.js
@@ -123,13 +123,15 @@ $(document).ready(function() {
                     command === 'delbookmark'
                 ) {
                     Object.keys(localStorage).forEach(function(key) {
-                        if (key.startsWith('bookmark.')) {
-                            const name = key.substr(9);
-                            choices.push({
-                                'label': `Bookmark: ${name}`,
-                                'value': `${command} ${name}`,
-                            });
+                        if (!key.startsWith('bookmark.')) {
+                            return;
                         }
+                        
+                        const name = key.substr(9);
+                        choices.push({
+                            'label': `Bookmark: ${name}`,
+                            'value': `${command} ${name}`,
+                        });
                     });
                 }
                 if (
--- a/serveradmin/servershell/static/js/servershell/autocomplete/command.js
+++ b/serveradmin/servershell/static/js/servershell/autocomplete/command.js
@@ -118,6 +118,21 @@ $(document).ready(function() {
                     });
                 }
                 if (
+                    command === 'bookmark' ||
+                    command === 'loadbookmark' ||
+                    command === 'delbookmark'
+                ) {
+                    Object.keys(localStorage).forEach(function(key) {
+                        if (key.startsWith('bookmark.')) {
+                            const name = key.substr(9);
+                            choices.push({
+                                'label': `Bookmark: ${name}`,
+                                'value': `${command} ${name}`,
+                            });
+                        }
+                    });
+                }
+                if (
                     command === 'orderby' ||
                     command === 'delattr' ||
                     command === 'history'
--- a/serveradmin/servershell/static/js/servershell/command.js
+++ b/serveradmin/servershell/static/js/servershell/command.js
@@ -235,6 +235,31 @@ servershell.commands = {
         // Now trigger reload with all changes at once
         servershell.shown_attributes = shown_attributes;
     },
+    bookmark: function(name) {
+        if (!name) {
+            servershell.alert('Missing name for bookmark', 'warning');
+            return;
+        }
+
+        localStorage.setItem('bookmark.' + name, servershell.shown_attributes);
+        servershell.alert(`Bookmark ${name} saved.`, 'success');
+    },
+    delbookmark: function(name) {
+        if (!name) {
+            servershell.alert('Missing name for bookmark', 'warning');
+            return;
+        }
+
+        localStorage.removeItem('bookmark.' + name);
+        servershell.alert(`Bookmark ${name} deleted.`, 'success');
+    },
+    loadbookmark: function(name) {
+        let attribute_ids = localStorage.getItem('bookmark.' + name);
+        if (attribute_ids) {
+            servershell.shown_attributes = attribute_ids.split(',');
+            servershell.alert(`Bookmark ${name} loaded.`, 'success');
+        }
+    },
     orderby: function(attribute_id) {
         if (!servershell.attributes.find(a => a.attribute_id === attribute_id)) {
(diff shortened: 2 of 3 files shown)


serveradmin/servershell/static/js/servershell/command.js
```diff
@@ -235,6 +235,17 @@ servershell.commands = {
         // Now trigger reload with all changes at once
         servershell.shown_attributes = shown_attributes;
     },
+    bookmark: function(name) {
+        let attribute_ids = localStorage.getItem(name);
+
+        if (attribute_ids) {
+            servershell.shown_attributes = attribute_ids.split(',');
+        }
+        else {
```
TheFRedFox, line 244: How do you overwrite a named bookmark?

serveradmin/servershell/static/js/servershell/autocomplete/command.js
```diff
@@ -117,6 +117,16 @@ $(document).ready(function() {
                         })
                     });
                 }
+                if (command === 'bookmark') {
+                    for (const key in localStorage) {
+                        if (localStorage.hasOwnProperty(key)) {
```
TheFRedFox, line 122: Is this if necessary? Shouldn't `const key in localStorage` already give you just properties?

TheFRedFox (commented):
2 discussion points.

serveradmin/servershell/static/js/servershell/command.js
kofrezo: You don't: feature not implemented yet.

serveradmin/servershell/static/js/servershell/autocomplete/command.js
kofrezo: Unfortunately yes. Otherwise you get also the object properties such as "length" etc.

TheFRedFox: I see .. you might want to try this:
``` js
Object.keys(localStorage)
```
Found it here: https://stackoverflow.com/a/17748203/2625488

serveradmin/servershell/static/js/servershell/command.js
TheFRedFox: I see ^^ .. might be interesting to leave a comment about that then here.

serveradmin/servershell/static/js/servershell/autocomplete/command.js
kofrezo: Will do

kofrezo: Features have been added. There is now bookmark, bookmarkadd and bookmarkdel command.

serveradmin/servershell/static/js/servershell/command.js
kofrezo: Features have been added. There is now bookmark, bookmarkadd and bookmarkdel command.

serveradmin/servershell/static/js/servershell/autocomplete/command.js
kofrezo: Adjust to use the above method.

serveradmin/servershell/static/js/servershell/command.js
```diff
@@ -235,6 +235,31 @@ servershell.commands = {
         // Now trigger reload with all changes at once
         servershell.shown_attributes = shown_attributes;
     },
+    bookmarkadd: function(name) {
```
TheFRedFox, line 238: Just a suggestion. Wouldn't it make more sense if `bookmark` would add/set a bookmark and `bookmarkload` or `load_bookmark` or `loadbookmark` would load it?
Just about wording, when I talk about "bookmark" then I think about setting a bookmark and not checking one.

serveradmin/servershell/static/js/servershell/command.js
```diff
@@ -235,6 +235,31 @@ servershell.commands = {
...
+        if (!name) {
+            servershell.alert('Missing name for bookmark', 'warning');
+            return;
+        }
+
+        localStorage.setItem(name, servershell.shown_attributes);
```
TheFRedFox, line 244: I guess it would already make sense to avoid conflicts later in the localStorage to prefix them all with a constant:
``` js
let key = "bookmark-" + name`
```
When the localStorage will be used for something else, you don't end up in mixing the values.
Then you would need to check for the prefix though by traversing through the items.
It seems that some are also using it like that:
``` js
let bookmarks = localStorage.getItem("bookmarks")
bookmarks.add(newBookmark)
localStorage.setItem("bookmarks", JSON.stringify(bookmarks))
```
So, you use more complex data inside one key. With that you don't need to traverse anymore through the items of the localStorage but just directly fetch the item data you need (and need to traverse that then).

One of these I would already do, so it is future safe.

serveradmin/servershell/static/js/servershell/command.js
kofrezo: Sure why not I don't have a strong preference - if it is more intuitive for you :shrug: ... let me search and replace.

TheFRedFox: I'm not a heavy serveradmin user, so whatever works better for the user. Was just my thinking for it already. ^^

kofrezo: Lets not overthink it. It is already adjusted.

serveradmin/servershell/static/js/servershell/command.js
kofrezo: Hmm yeah that probably makes sense.

kofrezo: Added `bookmark.` prefix everywhere.

serveradmin/servershell/static/js/servershell/autocomplete/command.js
```diff
@@ -117,6 +117,21 @@ $(document).ready(function() {
...
+                    command === 'delbookmark'
+                ) {
+                    Object.keys(localStorage).forEach(function(key) {
+                        if (key.startsWith('bookmark.')) {
+                            const name = key.substr(9);
+                            choices.push({
+                                'label': `Bookmark: ${name}`,
+                                'value': `${command} ${name}`,
+                            });
+                        }
```
TheFRedFox, line 132: Would just use early return here. Just to reduce some indention for the main code.
```suggestion
                        if (!key.startsWith('bookmark.')) {
                            return
                        }
                        
                        const name = key.substr(9);
                        choices.push({
                            'label': `Bookmark: ${name}`,
                            'value': `${command} ${name}`,
                        });
```

TheFRedFox (approved):

TheFRedFox (approved):

PR #168, opened 2021-06-29
Open source the hosts autocompletion

Felixoid:
Here's an autocompletion function used internally at the company. As discussed with Axel, it makes sense to add it to the project.

### DIFF
--- a/zsh-completion/adminapi
+++ b/zsh-completion/adminapi
@@ -1,0 +1,60 @@
+# Copyright © 2021, InnoGames GmbH
+# Query for all hosts from serveradmin API
+# This file should be put to the directory in ${FPATH},
+#   Arch Linux, Fedora: /usr/share/zsh/site-functions
+#   Debian: /usr/share/zsh/vendor-functions
+#
+# Configuration:
+# _SERVERADMIN_HOSTS_ENABLED: if the completion should be enabled
+# _SERVERADMIN_HOSTS_QUERY: customize hosts list, string
+# _SERVERADMIN_HOSTS_TIMEOUT: customize serveradmin query timeout, integer, seconds
+# _SERVERADMIN_HOSTS_UPDATE: invalidate cache interval, integer, minutes
+# _SERVERADMIN_HOSTS_ZSTYLE: change autocompletion behavior for hosts, bool
+function _serveradmin_hosts() {
+  # Not enable if adminapi is not found
+  command -v adminapi 1>/dev/null || return
+  local _query=${_SERVERADMIN_HOSTS_QUERY:-'intern_ip=not(empty())'}
+  local _hosts_cache="${HOME}/.ssh/serveradmin_hosts_cache"
+  local _tmp="${_hosts_cache}.tmp"
+  local _lock="${_hosts_cache}.lock"
+  local _invalidate_minutes=${_SERVERADMIN_HOSTS_UPDATE:-1}
+  local _timeout="${_SERVERADMIN_HOSTS_TIMEOUT:-8}s"
+  if [ "${_hosts_cache}"(Nmm-${_invalidate_minutes}) ]; then
+    # File exists and newer then $_invalidate_minutes minutes
+    # Use "<" as zsh internal `cat` for not invoking extra process
+    < "${_hosts_cache}"
+    return
+  fi
+
+  if [ -f "${_hosts_cache}" ]; then
+    # File exists and older then $_invalidate_minutes minutes
+    # First return list, then update it in background
+    < "${_hosts_cache}"
+    if   flock -n "${_lock}" true && [ -s "${_tmp}" ]; then
+      # If lockfile not locked and new temp cache already exists and not empty
+      mv "${_tmp}" "${_hosts_cache}"
+    elif flock -n "${_lock}" true; then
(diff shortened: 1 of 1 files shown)


adminapi/zsh-completion/adminapi
```diff
@@ -0,0 +1,62 @@
...
+# Configuration:
+# _SERVERADMIN_HOSTS_ENABLED: if the completion should be enabled
+# _SERVERADMIN_HOSTS_QUERY: customize hosts list, string
+# _SERVERADMIN_HOSTS_TIMEOUT: customize serveradmin query timeout, integer, seconds
+# _SERVERADMIN_HOSTS_UPDATE: invalidate cache interval, integer, minutes
+# _SERVERADMIN_HOSTS_ZSTYLE: change autocompletion behavior for hosts, bool
+function _serveradmin_hosts() {
+  # Not enable if adminapi is not found
+  command -v adminapi 1>/dev/null || return
+  local _query=${_SERVERADMIN_HOSTS_QUERY:-'servertype=any(blade_enclosure
```
kofrezo, line 16: The servertypes are specific to our data model. Can we get rid of them ?

adminapi/zsh-completion/adminapi
```diff
@@ -0,0 +1,62 @@
...
+    elif flock -n "${_lock}" true; then
+      # It shouldn't hang forever in case of problems with serveradmin
+      flock -n "${_lock}" timeout $_timeout \
+        adminapi $_query 2>/dev/null > "${_tmp}" &
+    fi
+    return
+  fi
+
+  # If file doesn't exists, then waiting for getting fresh
+  timeout $_timeout adminapi $_query 2>/dev/null | tee "${_hosts_cache}"
```
kofrezo, line 47: Out of curiosity: Have you tested how this ends the HTTP connection if we run into the timeout ? Will this be visible as 499 on the backend side ?

kofrezo (changes requested):
Lets discuss if we can get rid of the servertypes specific to our data model.

adminapi/zsh-completion/adminapi
Felixoid: Can you give me a default server types list, maybe?

kofrezo: There are none. We ship with no default data model. What is the reason you specified them in the first place at all - performance ?

adminapi/zsh-completion/adminapi
Felixoid: Yes, this is ended up as 499. But in our case, 8sec is sufficient timeout

kofrezo: Alright - sounds fair. Yes 8 seconds should be sufficient for simple Queries even for our size of database. This might just explain some 499 requests we have so it is good to know this can happen.

adminapi/zsh-completion/adminapi
Felixoid: this is the list of hosts, that will be used for autocompletion. It doesn't make sense to omit it completely, so the list reflects the internal needs.

kofrezo: I don't like the hard coded list because it renders it useless for anybody else unless you know you need to adjust it. Can we maybe provide this information via a configuration file in the home directory or environment variable and include it optionally ?

Felixoid: But it's the default, not the hardcoded query. One can define `_SERVERADMIN_HOSTS_QUERY`, then it will be used instead of the default value after `:-` mark.

This is from bash manual, behavior in zsh is the same:

```
${parameter:-word}
       Use Default Values.  If parameter is unset or null, the expansion of word is substituted.  Otherwise, the value of parameter is substituted.
```

lamaral: We can set the default to `not(empty(intern_ip))`. I think it's a good enough example.

Felixoid: > We can set the default to not(empty(intern_ip)). I think it's a good enough example.

Sounds good. It looks reasonable  
`adminapi 'intern_ip=not(empty())'  0.43s user 0.08s system 19% cpu 2.628 total`

kofrezo: Yes that sounds good.

kofrezo:
Does it make sense to move the autocompletion file to the adminapi module (it will not be shipped distributed with the Python module) or does it make more sense to move it to the root directory of the project so that people can locate it easier and setup their client or distribution package. 

I suggest moving it to `/zsh-autocompletion/adminapi`.

Felixoid:
> Does it make sense to move the autocompletion file to the adminapi module (it will not be shipped distributed with the Python module) or does it make more sense to move it to the root directory of the project so that people can locate it easier and setup their client or distribution package.
> 
> I suggest moving it to `/zsh-autocompletion/adminapi`.

Yes, it makes sense, done so

kofrezo (approved):

kofrezo:
No further remarks then. Thanks for your contribution @Felixoid - they will always be welcome :-)

Felixoid:
You are very welcome, I'll still watch for a couple of IG projects =)

PR #178, opened 2021-08-03
Fix new alert toggles previous one

kofrezo:
Make javascript alerts stacking so that we don't loose information.

### DIFF
--- a/serveradmin/servershell/static/js/servershell.js
+++ b/serveradmin/servershell/static/js/servershell.js
@@ -208,21 +208,21 @@ servershell.alert = function(text, level, auto_dismiss=5) {
         'dark'
     ];
 
-    if (!levels.includes(level))
+    if (!levels.includes(level)) {
         return;
+    }
 
-    let box = $('#js-alert');
-    levels.forEach(level => box.removeClass(`alert-${level}`));
-    box.addClass(`alert-${level}`);
+    let template = $('.alert').first();
 
-    let message = $('#js-alert-message');
-    message.text(text);
-
-    box.toggle();
+    let new_alert = template.clone();
+    new_alert.addClass(`alert-${level}`);
+    new_alert.children('.alert-text').text(text);
+    template.after(new_alert);
+    new_alert.toggle();
 
     if (auto_dismiss > 0) {
-        setTimeout(function () {
-            box.css('display', 'none');
+        setTimeout(function() {
+            $(new_alert).remove();
         }, auto_dismiss * 1000);
     }
 };
--- a/serveradmin/servershell/static/js/servershell.js
+++ b/serveradmin/servershell/static/js/servershell.js
@@ -212,17 +212,19 @@ servershell.alert = function(text, level, auto_dismiss=5) {
         return;
     }
 
-    let template = $('.alert').first();
+    let template = $('#alert-template');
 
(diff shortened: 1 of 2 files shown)


serveradmin/servershell/static/js/servershell.js
```diff
@@ -208,21 +208,21 @@ servershell.alert = function(text, level, auto_dismiss=5) {
...
         return;
+    }
 
-    let box = $('#js-alert');
-    levels.forEach(level => box.removeClass(`alert-${level}`));
-    box.addClass(`alert-${level}`);
-
-    let message = $('#js-alert-message');
-    message.text(text);
+    let template = $('.alert').first();
```
TheFRedFox, line 215: Let's give that template an id. There should be just one template and id-search is faster than class search plus first.

Even better would be to have it as a template string inside js and then replace variables, but for that you either need an extra library or ugly string replacement. So, just leaving it as an information here, not for this PR.

serveradmin/servershell/static/js/servershell.js
```diff
@@ -208,21 +208,21 @@ servershell.alert = function(text, level, auto_dismiss=5) {
...
 
-    box.toggle();
+    let new_alert = template.clone();
+    new_alert.addClass(`alert-${level}`);
+    new_alert.children('.alert-text').text(text);
+    template.after(new_alert);
```
TheFRedFox, line 220: Let's group all the alerts in an extra `div` with a class/id which you can reuse here and append on that group instead of putting it after the template. I would suggest to separate the template too, so maybe:
``` html
<div.alert-group#alert-group>
    <div.alert-template#alert-template />
    <div.alerts#alerts>
        <div.alert />
    </div>
</div>
```
Then you can find the template with `$('#alert-template')` and add new alerts (without knowing the exact method right now) with `$('#alerts').append(alert)`.

TheFRedFox (approved):
Strongly suggest to go for a bit structure change. Early approve though.

serveradmin/servershell/static/js/servershell.js
kofrezo: Using an id instead of an class would require to change/remove the id after cloning which imho reduces readibility and is also another function call. Since we are talking about microseconds here I am not sure if it is worth.

I don't like template strings in JS as the reduce readbility and make it hard for people not knowing the code to get a grasp of the page layout when looking at it. Do you think it has any advantages to make it a JS template string ?

serveradmin/servershell/static/js/servershell.js
kofrezo: Why is that better ?

kofrezo: Clarified in a call. That makes because the position of the template should not matter and this way we can append the new alert within the box. I will adjust the code.

serveradmin/servershell/static/js/servershell.js
kofrezo: Clarified in the call. 

Encapsulating this in a JS libary for reusability and scalability is not necessary here in my opinion because the Servershell page is the only one using this functionality (client side notifications). Everything else should use the Django messaging framework (server side) and use the alert template from base.html. This should always be our first choice.

PR #184, opened 2021-11-05
Provide development environment with docker-compose

kofrezo:
Make it easier to get local development started and ensure everbody has the same environment and dependencies.

### DIFF
--- a/.env.dist
+++ b/.env.dist
@@ -1,0 +1,14 @@
+# db service database settings
+POSTGRES_VERSION=14
+POSTGRES_DB=serveradmin
+POSTGRES_USER=serveradmin
+POSTGRES_PASSWORD=serveradmin
+
+# Credentials for default super user (requires Django >= 3.x)
+DJANGO_SUPERUSER_USERNAME=serveradmin
+DJANGO_SUPERUSER_EMAIL=serveradmin@exaple.com
+DJANGO_SUPERUSER_PASSWORD=serveradmin
+
+# adminapi credentials (python -m adminapi "example.com")
+SERVERADMIN_BASE_URL=http://127.0.0.1:8000/api
+SERVERADMIN_TOKEN=8igR21kG0yspOm9KRwUwozlu
--- a/.env.dist
+++ b/.env.dist
@@ -17,7 +17,7 @@ SERVERADMIN_TOKEN_OWNER=serveradmin
 SERVERADMIN_TOKEN_SUPERUSER=yes
 
 # SSH Host used by dump.sh
-PRODUCTION_DB=your-production-host.example.com
+REMOTE_DB=your-production-host.example.com
 
 # Django settings
 SECRET_KEY='DO_NOT_USE_IN_PRODUCTION_!!111'
--- a/.env.dist
+++ b/.env.dist
@@ -22,3 +22,4 @@ PRODUCTION_DB=your-production-host.example.com
 # Django settings
 SECRET_KEY='DO_NOT_USE_IN_PRODUCTION_!!111'
 DEBUG=True
+ALLOWED_HOSTS=127.0.0.1
--- a/.env.dist
+++ b/.env.dist
@@ -3,6 +3,7 @@ POSTGRES_VERSION=14
 POSTGRES_DB=serveradmin
 POSTGRES_USER=serveradmin
 POSTGRES_PASSWORD=serveradmin
+POSTGRES_HOST=db
 
 # Credentials for default super user (requires Django >= 3.x)
 DJANGO_SUPERUSER_USERNAME=serveradmin
@@ -15,3 +16,7 @@ SERVERADMIN_TOKEN=8igR21kG0yspOm9KRwUwozlu
 
 # SSH Host used by dump.sh
 PRODUCTION_DB=your-production-host.example.com
+
+# Django settings
+SECRET_KEY='DO_NOT_USE_IN_PRODUCTION_!!111'
+DEBUG=True
--- a/.env.dist
+++ b/.env.dist
@@ -7,7 +7,7 @@ POSTGRES_HOST=db
 
 # Credentials for default super user (requires Django >= 3.x)
 DJANGO_SUPERUSER_USERNAME=serveradmin
-DJANGO_SUPERUSER_EMAIL=serveradmin@exaple.com
+DJANGO_SUPERUSER_EMAIL=serveradmin@example.com
 DJANGO_SUPERUSER_PASSWORD=serveradmin
 
 # adminapi credentials (python -m adminapi "example.com")
--- a/.env.dist
+++ b/.env.dist
@@ -13,6 +13,8 @@ DJANGO_SUPERUSER_PASSWORD=serveradmin
 # adminapi credentials (python -m adminapi "example.com")
 SERVERADMIN_BASE_URL=http://127.0.0.1:8000/api
 SERVERADMIN_TOKEN=8igR21kG0yspOm9KRwUwozlu
+SERVERADMIN_TOKEN_OWNER=serveradmin
+SERVERADMIN_TOKEN_SUPERUSER=yes
 
 # SSH Host used by dump.sh
 PRODUCTION_DB=your-production-host.example.com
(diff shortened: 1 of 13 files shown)


serveradmin/common/management/commands/createdefaultuser.py
```diff
@@ -0,0 +1,31 @@
...
+    help = 'Create default user based on environment variables'
+
+    def handle(self, *args, **options):
+        username = environ.get('DJANGO_SUPERUSER_USERNAME')
+        password = environ.get('DJANGO_SUPERUSER_PASSWORD')
+
+        if not all([username, password]):
+            raise CommandError('Missing at least one DJANG_SUPERUSER_* env')
+
+        if not User.objects.filter(username=username).exists():
```
TheFRedFox, line 25: If the user already exists, should that command maybe update the existing user to make sure that the user is a staff and superuser?

serveradmin/apps/management/commands/createapp.py
```diff
@@ -0,0 +1,60 @@
...
+
+    def add_arguments(self, parser):
+        parser.add_argument('--owner', help='Username of application owner')
+        parser.add_argument('--superuser', action='store_true',
+                            help='Create superuser token')
+        parser.add_argument('--non-interactive', action='store_true', help=(
+            'Gather values from SERVERADMIN_TOKEN_OWNER, SERVERADMIN_TOKEN '
+            'and SERVERADMIN_TOKEN_SUPERUSER env variables. '
+            'Any value for SERVERADMIN_TOKEN_SUPERUSER means yes. '
+            'Empty string or absence means no.'
```
TheFRedFox, line 26: `None`, `empty list`, everything falseable means no. ^^

serveradmin/apps/management/commands/createapp.py
```diff
@@ -0,0 +1,60 @@
...
+                owner = input('Username of application owner (must exist!): ')
+            token = input('Token (empty for secure auto generated one): ')
+            superuser = options['superuser']
+
+        user = User.objects.filter(username=owner)
+        if not user.exists():
+            raise CommandError(f'No such user {owner} found!')
+
+        app = Application.objects.filter(
+            owner=user.get(), name=f'default app for {owner}')
```
TheFRedFox, line 48: In general (might be out of scope): If this is our indication of a default app (which we should also change in general ^^), then it would be good to have a central point of getting that specific name, so typos and change of the string is no problem in the future.

serveradmin/apps/management/commands/createapp.py
```diff
@@ -0,0 +1,60 @@
...
+        if app.exists():
+            self.stdout.write(self.style.WARNING(
+                f'Default app for {owner} already exists - skipping.'))
+            return
+
+        app = Application()
+        app.name = f'default app for {owner}'
+        app.owner = user.get()
+        app.superuser = superuser
+        if token != '':
```
TheFRedFox, line 58: Why comparison with `''` and not just `if not token`?

serveradmin/settings.py
```diff
@@ -17,6 +17,8 @@
 
 DEBUG = env('DEBUG', default=False)
 
+ALLOWED_HOSTS = env('ALLOWED_HOSTS', default=[])
```
TheFRedFox, line 20: `ALLOWED_HOSTS` in env is a string, but in python it needs to be a list. How is that working?

serveradmin/common/management/commands/createdefaultuser.py
kofrezo: No it should not for sake of security. The commands intention is to be used to setup an development environment and not override or update existing users.

serveradmin/apps/management/commands/createapp.py
kofrezo: There is no None, empty list etc. in shell if you define a variable your can write `SERVERADMIN_TOKEN_SUPERUSER=yes` or `SERVERADMIN_TOKEN_SUPERUSER=` or not define it. First one will always evaluate to true because it is a string > 0 later two will mark absence and therefore evaluate to False.

serveradmin/apps/management/commands/createapp.py
kofrezo: Not really. This is just for creating the development user and having a common string we can match to find the "default" application that should get created for the development environment.

serveradmin/apps/management/commands/createapp.py
kofrezo: No particular reason other than this is what I expect when the user does not specify a token. Why is `if not token` better ?

serveradmin/settings.py
kofrezo: I don't know you would have to check the implemention of the library or Django. 

I tested it with `ALLOWED_HOSTS="0.0.0.0"` (see .env.dists) and it seems to properly allow the correct hosts.

.env.dist
```diff
@@ -12,3 +12,6 @@ DJANGO_SUPERUSER_PASSWORD=serveradmin
 # adminapi credentials (python -m adminapi "example.com")
 SERVERADMIN_BASE_URL=http://127.0.0.1:8000/api
 SERVERADMIN_TOKEN=8igR21kG0yspOm9KRwUwozlu
+
+# SSH Host used by dump.sh
+PRODUCTION_DB=your-production-host.example.com
```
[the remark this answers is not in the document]
kofrezo: I couldn't think of a better name. I am not happy with `REMOTE_DB` either but it is better than `PRODUCTION_DB` so lets go with it.

.env.dist
```diff
@@ -0,0 +1,14 @@
+# db service database settings
+POSTGRES_VERSION=14
+POSTGRES_DB=serveradmin
+POSTGRES_USER=serveradmin
+POSTGRES_PASSWORD=serveradmin
+
+# Credentials for default super user (requires Django >= 3.x)
+DJANGO_SUPERUSER_USERNAME=serveradmin
+DJANGO_SUPERUSER_EMAIL=serveradmin@exaple.com
```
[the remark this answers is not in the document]
kofrezo: Fixed

serveradmin/apps/management/commands/createapp.py
TheFRedFox: True that.

serveradmin/apps/management/commands/createapp.py
TheFRedFox: `!= ''` will create `''` in memory and then checks the comparison between both memory values. `not token` does a direct falseable check on the `token` memory value.
So, 1. it is faster:
```
>>> setup = "tmp = ''"
>>> code = '''not tmp'''
>>> timeit.timeit(stmt=code, setup=setup)
0.024858662999974968
>>> code = "tmp != ''"
>>> timeit.timeit(stmt=code, setup=setup)
0.03584977599996364
```
2. In general the `not token`, so the falsy check is also more pythonic to use. Took me a while to find the statement but here: https://stackoverflow.com/a/9573283/2625488 (in the comment PEP8 ist getting citated)

serveradmin/apps/management/commands/createapp.py
TheFRedFox: So, this name is just the name of the scope of this exact command here? Then nevermind. A const might be interesting though to not duplicate the string in line 48 and 55, but totally minor.

serveradmin/settings.py
TheFRedFox: Most probably that's then some magic of the `django-environ` library.

.env.dist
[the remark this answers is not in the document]
TheFRedFox: Yeah, I know, I also think `REMOTE_DB` is still wrong, but can't think of a better one. `DUMPING_DB`? ^^ `COPY_FROM_DB`?

serveradmin/apps/management/commands/createapp.py
kofrezo: Okay then. I'll grab the tip from stackoverflow and strip the token then beforehand just to be sure nobody tries to create a token consisting of empty string.

TheFRedFox: Actually, the comment about `strip` strings is the comment after the comment I meant. ^^ But, yes, it might be better to first strip the string to not allow tokens consisting just of whitespace.

TheFRedFox (approved):

kofrezo:
Just waiting for a few more people (internally) to test the setup and confirm that it "works on their machine".

PR #185, opened 2021-11-05
Open documentation in a new tab

kofrezo:


kofrezo:
Tests are failing because of a missing command sphinx-build for the docs. This PR should be safe though.

PR #186, opened 2021-11-05
Update Pipfile dependencies to latest versions

kofrezo:


kofrezo:
Tests are failing because docs cannot be build because of a missing sphinx-build command.

PR #189, opened 2021-11-11
Be able to exclude changes matching a string

kofrezo:
It can be difficult to view the history of Serveradmin objects where
certain attributes are frequently changed (e.g. by a script).

Until the history becomes a jsonb type in PostgreSQL and we can filter
for specific attributes we work around this by matching strings.


TheFRedFox (approved):
LGTM

PR #191, opened 2021-11-15
Dismiss alerts by <ESC> keyboard shortcut

kofrezo:
Using the mouse to dismiss alerts can be tedious and often one doesn't
want to wait n seconds before the alert disappears.

### DIFF
--- a/serveradmin/common/static/js/plugins/esc_dismissible.js
+++ b/serveradmin/common/static/js/plugins/esc_dismissible.js
@@ -1,0 +1,31 @@
+/*
+ * Dismiss by ESC keyboard shortcut, Copyright (c) 2021 InnoGames GmbH
+ *
+ * Allows to dismiss (delete) HTML elements that have the class dismissible
+ * by pressing escape.
+ */
+$(document).ready(function() {
+    $(document).keydown(function (event) {
+        // ESC
+        if (event.key === "Escape") {
+            let dismissible = $('.dismissible');
+
+            if (dismissible.length < 1) {
+                return;
+            }
+
+            // Pressing ESC on elements such as inputs makes them loose focus.
+            // To avoid this we remember the focused element to restore it.
+            let input = document.activeElement;
+
+            dismissible.remove();
+
+            if (input) {
+                input.focus();
+            }
+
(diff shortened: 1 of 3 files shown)


serveradmin/common/static/js/plugins/esc_dismissible.js
```diff
@@ -0,0 +1,32 @@
...
+
+            dismissible.remove();
+
+            if (input) {
+                input.focus();
+            }
+
+            event.preventDefault();
+            event.stopPropagation();
+            return false;
```
TheFRedFox, line 29: Could be interesting to know if really both is needed or actually just one of it.

serveradmin/common/templates/base.html
```diff
@@ -79,6 +79,7 @@
 <script src="{{ STATIC_URL }}js/js.cookie.min.js"></script>
 <script src="{{ STATIC_URL }}js/serveradmin.js"></script>
 <script src="{{ STATIC_URL }}js/plugins/terminal_keyboard.js"></script>
+<script src="{{ STATIC_URL }}js/plugins/esc_dismissible.js"></script>
```
TheFRedFox, line 82: I wonder if it would make more sense to put it into the `servershell.js`.

TheFRedFox (approved):
Just 2 minor discussions.

serveradmin/common/templates/base.html
kofrezo: I first put it to the servershell app but then decided that it could be handy to be able to use this functionality in other apps as well. This allows us to just add a CSS class dismissible to any HTML element and discard it with escape.

TheFRedFox: Fair enough! 👍🏼

serveradmin/common/static/js/plugins/esc_dismissible.js
kofrezo: Indeed the `return false` seems to be redundant. The MDN Web Docs don't mention `return false` on the pages of `preventDefault` or `stopPropagation` as alternative to achieve the same so I prefer them over the `return false` and remove it.

PR #195, opened 2021-11-22
Expand gitignore

seqizz:
Managed to run the whole environment via nix-shell. These 2 are globally useless for git.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -73,3 +73,7 @@ __pycache__/
 
 # Virtual Environment
 /venv/
+
+# Nix-related
+/.envrc
+/default.nix


kofrezo:
As these files are specific to a operating system and not Python, Django or its build tools I think the should not go to the `.gitignore` of the project but rather the `.gitignore` of the host:

> Patterns which a user wants Git to ignore in all situations (e.g., backup or temporary files generated by the user’s editor of choice) generally go into a file specified by core.excludesFile in the user’s ~/.gitconfig. Its default value is $XDG_CONFIG_HOME/git/ignore. If $XDG_CONFIG_HOME is either not set or empty, $HOME/.config/git/ignore is used instead.

See https://git-scm.com/docs/gitignore

What do you think ?

seqizz:
Hmm I could argue this is no different than .eggs, since it's a way of packaging.
But didn't know about the global gitignore :exploding_head: With that option, feel free to merge or close the PR tbh. I'll set this up there :+1:

PR #196, opened 2021-11-22
Add PowerDNS database and nameserver service

kofrezo:
Extend docker-compose development setup for upcoming PowerDNS integration.


kofrezo:
Credentials here are generic credentials for the development environment necessary (PostgreSQL docker container cannot be started without any) and therfore save to check into VCS.

PR #200, opened 2021-12-03
Include powerdns templates in package_data

kofrezo:


kofrezo:
Forgot to include this in!199

PR #201, opened 2021-12-03
Add Domain and Record models

kofrezo:
Add not managed models to access domains and records in the powerdns database.

### DIFF
--- a/serveradmin/powerdns/models.py
+++ b/serveradmin/powerdns/models.py
@@ -5,6 +5,8 @@ class Domain(models.Model):
     """PowerDNS Domain
 
     Model to access domain in the PowerDNS db.
+
+    See https://doc.powerdns.com/authoritative/backends/generic-postgresql.html#default-schema
     """
     TYPES = [
         ('MASTER', 'MASTER'),
@@ -26,6 +28,8 @@ class Record(models.Model):
     """PowerDNS Record
 
     Model to access records in the PowerDNS db.
+
+    See https://doc.powerdns.com/authoritative/backends/generic-postgresql.html#default-schema
     """
     TYPES = [
         ('A', 'A'),
--- a/serveradmin/powerdns/models.py
+++ b/serveradmin/powerdns/models.py
@@ -1,3 +1,22 @@
 from django.db import models
 
-# Create your models here.
+
+class Domain(models.Model):
+    """PowerDNS Domain
+
+    Model to access domain in the PowerDNS db.
+    """
+    TYPES = [
+        ('MASTER', 'MASTER'),
+        ('SLAVE', 'SLAVE'),
+        ('NATIVE', 'NATIVE'),
+    ]
+
+    id = models.IntegerField(primary_key=True)
+    name = models.CharField(max_length=255, null=False)
+    master = models.CharField(max_length=128, default=None)
+    type = models.CharField(max_length=6, null=False, choices=TYPES)
+
+    class Meta:
+        managed = False
+        db_table = 'domains'
--- a/serveradmin/powerdns/models.py
+++ b/serveradmin/powerdns/models.py
@@ -20,3 +20,32 @@ class Domain(models.Model):
     class Meta:
         managed = False
         db_table = 'domains'
+
+
+class Record(models.Model):
+    """PowerDNS Record
+
+    Model to access records in the PowerDNS db.
+    """
+    TYPES = [
+        ('A', 'A'),
+        ('AAAA', 'AAAA'),
+        ('CNAME', 'CNAME'),
+        ('TXT', 'TXT'),
+        ('SSHFP', 'SSHFP'),
+        ('SOA', 'SOA'),
+        ('MX', 'MX'),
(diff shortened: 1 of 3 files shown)


serveradmin/powerdns/models.py
```diff
@@ -1,3 +1,22 @@
...
+    """
+    TYPES = [
+        ('MASTER', 'MASTER'),
+        ('SLAVE', 'SLAVE'),
+        ('NATIVE', 'SLAVE'),
+    ]
+
+    id = models.IntegerField(primary_key=True)
+    name = models.CharField(max_length=255, null=False)
+    master = models.CharField(max_length=128, default=None)
```
TheFRedFox, line 17: What will be stored in here? Thought it would be a reference to the master then if it is a slave, but then I would assume a ForeignKey to Domain itself.

serveradmin/powerdns/models.py
```diff
@@ -1,3 +1,22 @@
...
+
+class Domain(models.Model):
+    """PowerDNS Domain
+
+    Model to access domain in the PowerDNS db.
+    """
+    TYPES = [
+        ('MASTER', 'MASTER'),
+        ('SLAVE', 'SLAVE'),
+        ('NATIVE', 'SLAVE'),
```
TheFRedFox, line 12: There is a movement in Software Development to go away from Master/Slave wording to Primary/Secondary or similar (https://en.wikipedia.org/wiki/Master/slave_(technology)#Terminology_concerns). Is it for this purpose here also possible?

serveradmin/powerdns/models.py
```diff
@@ -20,3 +20,32 @@ class Domain(models.Model):
...
+        ('MX', 'MX'),
+        ('PTR', 'PTR'),
+        ('NS', 'NS'),
+    ]
+
+    id = models.BigIntegerField(primary_key=True)
+    domain_id = models.ForeignKey('Domain', on_delete=models.CASCADE)
+    name = models.CharField(max_length=255, default=None)
+    type = models.CharField(max_length=10, default=None, choices=TYPES)
+    content = models.CharField(max_length=65535, default=None)
```
TheFRedFox, line 46: If it is that big, might it be better to directly use a `text` field?

serveradmin/powerdns/routers.py
```diff
@@ -0,0 +1,22 @@
+class PowerDNSRouter:
```
TheFRedFox, line 1: If such routers are needed more often, I can see a possibility to provide something which would make adding such easier. So, just already a hint to poke me, if it is needed. ;)

serveradmin/settings.py
```diff
@@ -54,6 +54,7 @@
         },
     },
 }
+DATABASE_ROUTERS = ['serveradmin.powerdns.routers.PowerDNSRouter']
```
TheFRedFox, line 58: Weren't there more Routers already? Thought, I remember that we already added some routers before. But maybe I misremembering something here.

TheFRedFox (commented):
Some discussion points, nothing major. Will approve once some discussion points are resolved.

serveradmin/powerdns/models.py
kofrezo: I don't know yet. It is a given field of the PowerDNS [PostgreSQL schema](https://doc.powerdns.com/authoritative/backends/generic-postgresql.html#default-schema).

In our current database we don't use the field yet (it is null) so it might go away at a later point.

Maybe it is a good idea to link the schema for reference.

serveradmin/powerdns/models.py
kofrezo: No it is not. These are internal operation modes of PowerDNS. But theere is indeed a mistake second value in tuple with `NATIVE` should be `NATIVE`.

serveradmin/powerdns/routers.py
kofrezo: Nothing planned.

serveradmin/powerdns/models.py
TheFRedFox: 👍🏼  for the schema reference linking!

serveradmin/settings.py
kofrezo: No there are no more routers currently. Very early back in the days there may have been some when we were switching from MySQL to PostgreSQL but I am not sure.

serveradmin/powerdns/models.py
kofrezo: Given by the PowerDNS schema (mind the model is not managed).

serveradmin/settings.py
TheFRedFox: Ah, possible. Nevermind then.

TheFRedFox (approved):
Just a comment with a link/reference about the schema would be really good to have for later adjustments.
https://github.com/innogames/serveradmin/pull/201#discussion_r762011355

serveradmin/powerdns/models.py
kofrezo: I linked it.

PR #210, opened 2022-01-24
Add migrations for (datetime) attribute, servertype

kofrezo:
This changes seem to be missing from the time before we included
migrations.


kofrezo:
When applying the changes to our current database they don't apply any changes (DDL statements) so I assume this is just a left over from the time before we added migrations.

PR #215, opened 2022-02-11
Create PowerDNS domains for created objects

kofrezo:


kofrezo:
- Will add some tests here
- Using the `powerdns` branch to experiment around with some implementetions

PR #219, opened 2022-02-17
Add missing class and method super calls

kofrezo:


serveradmin/serverdb/tests/test_ip_addr_type.py
```diff
@@ -27,9 +27,11 @@ class TestIpAddrType(TransactionTestCase):
 
     @classmethod
     def setUpClass(cls):
+        super(TestIpAddrType, cls).setUpClass()
```
Woellchen, line 30: are the `super()` calls with parameters necessary here? it should already be the default no?

kofrezo: You were too late Timo was faster.

PR #221, opened 2022-02-24
Delete PowerDNS domains on object deletion

kofrezo:

PR #222, opened 2022-02-24
Test PowerDNS domain deletion

kofrezo:


kofrezo:
This belongs to #221. Forgot to add it.

PR #227, opened 2022-03-10
Update PowerDNS domains on post_commit signal

kofrezo:


kofrezo:
When executing the tests locally I get some warnings about tests accessing the pdns database even tough powerdns signals should be disabled. 

I want to look into this first to ensure our test setup is working properly.

kofrezo:
> When executing the tests locally I get some warnings about tests accessing the pdns database even tough powerdns signals should be disabled.
> 
> I want to look into this first to ensure our test setup is working properly.

This is caused because in the tests we import `serveradmin.powerdns.signals` to unit test the functions which triggers the `@receiever` decorator and that connects the signals. 

I'll make a dedicated PR where I change the way we connect the callbacks to the signals in a safe way.

PR #235, opened 2022-03-25
Servershell: sort multi attributes alphabetically

brainexe:
before
->
after


kofrezo:
:+1: I think sorting them in the fronted is the fitting and most simple solution here. This is likely a thing we only care in the Servershell for and not in the API itself.

PR #237, opened 2022-03-25
servershell: Add implicit attr/setattr/multiadd

YannikSc:
Adds shortcuts for attr/setattr/multiadd

## Example

Command: `attribute_name` interpreted as `attr attribute_name` -> Toggles the attributes column

Command: `attribute_name=some_value` interpreted as `setattr attribute_name=some_value` (multiadd if the attribute is multi=true) -> sets/adds a value to the attribute

### DIFF
--- a/serveradmin/servershell/static/js/servershell/command.js
+++ b/serveradmin/servershell/static/js/servershell/command.js
@@ -678,7 +678,7 @@ function execute_command() {
     }
 
     let [attribute_name, value] = servershell.command.split('=');
-    let attribute = servershell.attributes.find(attribute => attribute.attribute_id === attribute_name);
+    let attribute = servershell.attributes.find(attribute => attribute.attribute_id === attribute_name.trim());
 
     if (attribute && !value) {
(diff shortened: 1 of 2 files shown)


serveradmin/servershell/static/js/servershell/command.js
```diff
@@ -655,6 +655,51 @@ servershell.commands = {
     },
 };
 
+function execute_command() {
```
kofrezo, line 658: The indentation looks odd. Are these 4 spaces ?

kofrezo, line 658: Could you please add a doc string here describing what the function does.

kofrezo (changes requested):
Please double check the indentation and add a doc string.

The features works as expected but (of course) auto completion for the attributes does not work now anymore (e.g. typing resp will not give a suggestion for responsible_admin). 

Do you think you can easily get this working here too or should we make a dedicated PR for this ?

kofrezo (approved):

PR #241, opened 2022-03-25
Prevent duplicate ips in choose_ip addr and reduce number

zados:
When using Serveradmin for multi home setups, it might occur that the same network needs to be defined for two datacenters, which will result in duplicate IPs when using `choose_ip_addr`. 

Also the current count of 1000 is too high for the average user. 50 is more appropriate.


kofrezo:
If I understood correctly in PM @TuxPowered42 has some objections here ?

TuxPowered42:
> defined for two datacenters

A network of given type is present in Serveradmin *always* only once. It can't be in 2 datacenters, as the `datacenter` attrribute is a single-attribute. Being in multiple `network_zones` (a multi-atttribute) still would not change the behaviour. From the discussuion last week the conclusion was that networks of wrong type are queried and shown together. The IP address picker must present only one network type at a time. By IG convention this is first the `route_network` and once that is clicked, `project_network`. IP addresses of `provider_network` must never be shown to the user as they are only used by NDCO to document networks given by the provider.

> Also the current count of 1000 is too high for the average user. 50 is more appropriate.

This is a very good idea.

kofrezo (approved):

PR #242, opened 2022-03-25
Forbid creating graphite collections with duplicated params

v0devil:


kofrezo:
Can we add this to the Template, Variation and Numeric params input as well ?

serveradmin/graphite/admin.py
```diff
@@ -38,4 +38,16 @@ class CollectionAdmin(admin.ModelAdmin):
     list_filter = ['overview']
 
 
-admin.site.register(Collection, CollectionAdmin)
+class CollectionAdminForm(forms.ModelForm):
+
+    def clean_params(self):
```
kofrezo, line 43: Can we add this to the Template, Variation and Numeric params input as well ?

kofrezo: I am getting an `AttributeError` exception when submitting a parameter value for Numeric or Variation now:

```
'NumericFormFormSet' object has no attribute 'cleaned_data'
```

```
'VariationFormFormSet' object has no attribute 'cleaned_data'
```

v0devil: Ok, it cuz other form field were invalid

PR #243, opened 2022-03-25
Fix ContainedBy filter for IP networks

meikomeik:
Disables IP Network strict mode

### DIFF
--- a/adminapi/datatype.py
+++ b/adminapi/datatype.py
@@ -135,5 +135,9 @@ def json_to_datatype(value):
             # the outputs to match.
             if datatype is EUI:
                 return EUI(value, dialect=mac_unix_expanded)
+            # Disable strict mode for IP Networks because they might have
+            # host bits set. See ipaddress.IPv6Network module page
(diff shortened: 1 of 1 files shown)


TuxPowered42 (changes requested):
Let's not ignore the user's error and just overwrite the request. The proper way would be to report the error back to the UI and have the user type in proper data.

meikomeik:
I changed the behaviour to not ignore the "<ip> has host bits set" but rather raise it as a FilterValueError.
This way it will be displayed to the user in the serveradmin frontend

kofrezo:
The method `json_to_datatype` is currently only used by the backend in the Servershell (`servershell/results`) and not by the Python API (`api/dataset/qery`) - Probably for historic reasons.

This means the validation only takes place when using the Servershell. 

Refactoring the backend to use the same logic is beyond this PR and so we'll merge this as it already gives a benefit with having the validation in the backend.

kofrezo:
Example of the error message in the Servershell instead of a HTTP status code 500.

Screenshot from 2022-03-30 08-42-55

PR #246, opened 2022-03-25
"export" command: use all selected attributes by default

brainexe:
if no attributes are passed in the "export" command, use all selected ones in the current servershell
Screenshot 2022-03-25 at 17 11 52


TheFRedFox (approved):
LGTM! :)

PR #247, opened 2022-03-25
Remember attribute list status

seqizz:
Now we use localStorage to remember the state of attribute list. If it is unset, we don't change the default (shown).

### DIFF
--- a/serveradmin/servershell/static/js/servershell.js
+++ b/serveradmin/servershell/static/js/servershell.js
@@ -253,3 +253,23 @@ servershell.get_attribute = function(attribute_id) {
     return servershell.attributes.find(
         attribute => attribute.attribute_id === attribute_id);
 };
+
+/**
+ * Save the status of attributes list
+ */
+servershell.save_hidden_attrlist = function(element) {
+    localStorage.setItem('attrlist_hidden', (element.attributes["aria-expanded"]["nodeValue"] === 'true'));
+};
+
+/**
+ * Check the local storage and toggle attributes list
+ *
+ * If it's not defined or set to not hidden, it triggers collapse function
+ * to show. Otherwise it keeps it closed.
+ */
+servershell.sync_attrlist_visibility = function() {
+    let attribute_list_hidden = localStorage.getItem('attrlist_hidden');
+    if ( attribute_list_hidden === "false" || attribute_list_hidden === null ) {
+        $("#accordion-attributes-body").collapse('show');
+    };
+};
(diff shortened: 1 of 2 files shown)


serveradmin/servershell/static/js/servershell.js
```diff
@@ -252,4 +252,24 @@ servershell.get_object = function(object_id) {
...
+    localStorage.setItem('attrlist_hidden', (element.attributes["aria-expanded"]["nodeValue"] === 'true'));
+};
+
+/**
+ * Check the local storage and toggle attributes list
+ *
+ * If it's not defined or set to not hidden, it triggers collapse function
+ * to show. Otherwise it keeps it closed.
+ */
+servershell.set_attrlist_state = function() {
```
TheFRedFox, line 270: Why you chose `set_attrlist_state` is clear to me, however the `set` adds the connotation that you are able to set the state and not that it will automatically set.
Maybe `update_attrlist_visibility` or `sync_attrlist_visibility` might be a better choice of naming.

serveradmin/servershell/templates/servershell/index.html
```diff
@@ -105,9 +105,9 @@
                 <div class="card">
                     <div class="card-header">
                         <b>Attributes</b>
-                        <button class="btn btn-link" type="button" data-toggle="collapse" data-target="#accordion-attributes-body"></button>
+                        <button class="btn btn-link collapsed" type="button" data-toggle="collapse" aria-expanded="false" id="attributesCollapseButton" data-target="#accordion-attributes-body" onclick="servershell.save_hidden_attrlist(this)"></button>
```
TheFRedFox, line 108: When I look through the html lines here, I see snake_casing being used a lot for ids and other stuff. So, we should stick with it and not adding another casing:
```suggestion
                        <button class="btn btn-link collapsed" type="button" data-toggle="collapse" aria-expanded="false" id="attributes_collapse_button" data-target="#accordion-attributes-body" onclick="servershell.save_hidden_attrlist(this)"></button>
```
I would say, normally one use kebab-casing in html lines.

serveradmin/servershell/static/js/servershell.js
seqizz: Doing sync_attrlist_visibility

serveradmin/servershell/templates/servershell/index.html
seqizz: Realized we don't even need to give ID here anymore.

> use kebab-casing

I see what you did there :smirk:

kofrezo (approved):

PR #250, opened 2022-03-31
Fix Attribute was modified despite being whitelisted

kofrezo:
When a query tries to commit changes it sometimes fails because the
command to cache graphite numeric has changed an attribute of this
server in the meantime.

### DIFF
--- a/serveradmin/graphite/management/commands/cache_graphite.py
+++ b/serveradmin/graphite/management/commands/cache_graphite.py
@@ -5,6 +5,7 @@ Copyright (c) 2019 InnoGames GmbH
 
 import json
 from datetime import datetime
+from decimal import Decimal
 from os import mkdir
 from os.path import isdir
 import time
@@ -123,7 +124,7 @@ class Command(BaseCommand):
                 locked_server.servernumberattribute_set.update_or_create(
                     server_id=locked_server.server_id,
                     attribute=numeric.attribute,
-                    defaults={'value': value},
+                    defaults={'value': Decimal(value)},
                 )
 
     def get_from_graphite(self, params):
--- a/serveradmin/graphite/management/commands/cache_graphite.py
+++ b/serveradmin/graphite/management/commands/cache_graphite.py
@@ -4,24 +4,25 @@ Copyright (c) 2019 InnoGames GmbH
 """
 
 import json
+import time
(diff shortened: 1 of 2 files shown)


serveradmin/serverdb/query_committer.py
```diff
@@ -108,6 +119,8 @@ def commit_query(created=[], changed=[], deleted=[], app=None, user=None):
         _upsert_attributes(attribute_lookup, changed, changed_servers)
         changed_objects = _materialize(changed_servers, joined_attributes)
 
+        # @TODO Improve this function by checking only attributes of ACLs that
```
TheFRedFox, line 122: Here there is suddenly a `@` in front of the "TODO". Is that on purpose?

serveradmin/graphite/management/commands/cache_graphite.py
```diff
@@ -116,10 +117,14 @@ def cache_numerics(self, collection, server):
             # it is set up like this.  This process takes a long time.
             # We want the values to be immediately available to the users.
             with transaction.atomic():
-                ServerNumberAttribute.objects.update_or_create(
-                    server_id=server.object_id,
+                # Lock server for changes to avoid nonrepeatable reads in the
+                # query_committer.
+                locked_server = Server.objects.select_for_update().get(
```
TheFRedFox, line 122: Do I understand it correctly, that this works before it would always first try and lock the server and if not wait for it? If some code would not try and first lock the server but directly work on the attributes again it would again fail, right?
Just checking if I understood it correctly right now.

serveradmin/serverdb/query_committer.py
kofrezo: No it's just what my IDE does automatically. I removed it.

serveradmin/graphite/management/commands/cache_graphite.py
kofrezo: Yes exactly. It would wait for a lock (same as in the query_committer) instead of just updating the attribute possibliy causing a nonrepeatable read in the query committer.

And yes you are right it could happen again if any other code would not follow this "protocol". Hence the comment in the query committer to evaluate if we can change the transaction level to "repeatable read" so that it is robust against other code places.

PR #255, opened 2022-06-01
Ignore newlines in query string

kofrezo:
I can't think of a good reason to have newlines in a query string.
Removing them allows to type a query in the terminal across multiple
lines without changing the query.

### DIFF
--- a/adminapi/parse.py
+++ b/adminapi/parse.py
@@ -10,6 +10,9 @@ _trigger_re_chars = ('.*', '.+', '[', ']', '|', '\\', '$', '^', '<')
 
 
 def parse_query(term, hostname=None):  # NOQA: C901
+    # Ignore newlines to allow queries across multiple lines
+    term = term.replace('\n', '')
+
     parsed_args = parse_function_string(term, strict=True)
     if not parsed_args:
         return {}


kofrezo:
This will allow typing long queries across multiple lines without changing the query.

This for example leads to different results currently because of the newline.
```
#with spaces
$ adminapi 'servertype=any(blade_enclosure 
    floating_address hardware hw_internalrouter hw_loadbalancer hypervisor 
    loadbalancer natbox switch vm vm_external vpn)' | wc
   9234    9234  208610
#wothout spaces
$ adminapi 'servertype=any(blade_enclosure
    floating_address hardware hw_internalrouter hw_loadbalancer hypervisor
    loadbalancer natbox switch vm vm_external vpn)' | wc
   8751    8751  196718
```

adminapi/parse.py
```diff
@@ -10,6 +10,9 @@
 
 
 def parse_query(term, hostname=None):  # NOQA: C901
+    # Ignore newlines to allow queries across multiple lines
+    term = term.replace('\n', '')
```
TheFRedFox, line 14: Normally, in terminals or the like, new lines are seen as spaces, no? 
So, would it make sense to use that here too? To not replace a new line with nothing but with a space?
```suggestion
    term = term.replace('\n', ' ')
```

kofrezo: I know the behavior from HTML. I actually tested pasting a query separated by newline to the Servershell search input and it acted as you described.

While implementing it I thought about replacing it with a whitespace as you suggested but couldn't find any example where a terminal would behave like replacing a newline with a whitespace and therefore rather sticked with the above.

Thus my curiosity about an example.

PR #269, opened 2022-10-11
Fix ProgrammingError: syntax error near "AND"

kofrezo:
Remove extra comma in SQL query string concatention to avoid joining it together as 2 "where" conditions with AND.


Woellchen (approved):
LGTM

PR #297, opened 2023-03-08
Use internal_domain attribute to generate A, AAAA records

kofrezo:


kofrezo:
This allows us, with our servertype attribute model at InnoGames, to create A and AAAA records for internal DNS loadbalancing.

PR #308, opened 2023-04-04
Rename restore view to recreate for clarity

kofrezo:
We are not restoring the original object but create a new one with the same data.


kofrezo:
fixes bug introduced with #304

PR #323, opened 2023-07-21
Improve graphite cache generation

kofrezo:
- Reduce runtime by requesting sprites and numerics concurrently from Graphite
- Allow to update only specific collections for specific hosts
- Add details to logging output, print current time and add colors
- Don't raise exception if host has been deleted while generating sprites and numerics


kofrezo:
Note: Check if `wait(futures)` in commit running generation asynchronously to ensure all threads are done is needed.

PR #324, opened 2023-07-21
Fix minor UI bugs for resources page

kofrezo:
- Do not reset view selection when changing the page
- Load last viewed collection by default


kofrezo:
This branch currently contains the changes of #323. So that one should be merged first.

PR #340, opened 2024-01-30
Support ED22519 format for security keys

kofrezo:


kofrezo:
The patch should allow security key ed25519 keys but currently it fails in the [line](https://github.com/innogames/serveradmin/pull/340/files#diff-22b5fd159526f42b37370622bd524862f72d39af15aea9c261e86ffe0f5f83a7L107) afterwards because paramiko (tested 3.3.1) does not support security keys yet.
```
return Ed25519Key(data=public_key_blob)
```

See https://github.com/paramiko/paramiko/issues/2299

PR #343, opened 2024-02-05
Double signing speed by signing optimistically

Woellchen:
Skip the checks for usable signing keys and just try to sign the data


kofrezo:
I couldn't verify the speed improvement on my local computer it was roughly the same but since it is more simple it's worth it anyways.

PR #346, opened 2024-02-15
Add new attributes for separate IPv4 and IPv6 addresses

TuxPowered42:


kofrezo:
Add some commits as code review (suggestions) to discuss. We can drop them if not needed and I adjusted the tests for the new behaviour.

PR #356, opened 2024-04-30
Introduce ruff for codestyle and linting

kofrezo:


kofrezo:
Running this for some time on staging to ensure it breaks nothing.

PR #363, opened 2024-06-27
Retry signing the request if too slow

kofrezo:
There are cases where the cryptographic operation to sign the request takes unusually long. One example is when using a hardware such as the YubiKey which is busy signing other stuff.

The Serveradmin server is pretty strict and would reject these as expired for security reasons.

To gracefully handle temporary congestion in the signing we log an error and retry the operation up to Settings.tries.

### DIFF
--- a/adminapi/request.py
+++ b/adminapi/request.py
@@ -2,7 +2,7 @@
 
 Copyright (c) 2019 InnoGames GmbH
 """
-
+import logging
 import os
 from hashlib import sha1
 import hmac
@@ -41,6 +41,9 @@ from adminapi.exceptions import (
 )
 
 
+logger = logging.getLogger(__name__)
+
+
 def load_private_key_file(private_key_path):
     """Try to load a private ssh key from disk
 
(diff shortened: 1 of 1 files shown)


adminapi/request.py
```diff
@@ -184,6 +188,15 @@ def _build_request(endpoint, get_params, post_params):
...
         headers['X-Signatures'] = ','.join(key_signatures.values())
 
+    time_spent_signing = int(time.time()) - timestamp
+    if time_spent_signing > Settings.grace_period:
+        if retry <= Settings.tries:
+            logger.error(
+                f'Signing the requests took {time_spent_signing} seconds! '
+                'Serveradmin would reject this request. Maybe your signing '
+                f'soft-/hardware is congested ? Retry {retry}/{Settings.tries}.')
+            return _build_request(endpoint, get_params, post_params, retry+1)
```
kofrezo, line 198: Question to my reviewers: Should we consider a random sleep here ?

v0devil: I think for now we can leave it without sleeps

v0devil (approved):

PR #365, opened 2024-07-04
Implement adminapi in Rust

YannikSc:
This PR implements the adminapi in Rust supporting:

- SSH Authentication
- Token authentication
- Querying of objects
- Creation of new objects

Other APIs (like pushing the firewall) are not yet implemented, as these come from external modules


kofrezo:
Moved to a dedicated project https://github.com/innogames/adminapi-rs

PR #376, opened 2024-09-02
Downgrade minimum version for netaddr

kofrezo:
There are some internal libraries that need some changes first.


kofrezo:
I will bypass the code review here as it is only an adjustment in the dependencies necessary due to internal libraries.

PR #378, opened 2024-09-05
Fix pagination does not work on history page

kofrezo:
Set form_id parameter to ensure pagination is finding the correct form to submit.


brainexe (approved):
works fine locally!

PR #379, opened 2024-09-11
Render HTML contained in messages

kofrezo:
For some messages it is useful to render the HTML such as for example if you have multiple messages and want to display them as list.


lamaral (commented):
It LGTM.
I will wait to hear what @zados says, as I see a potential for XSS.

zados (approved):

PR #380, opened 2024-12-03
servershell: Add column helper buttons

seqizz:
Adds 2 small buttons next to each column which triggers orderby/attr commands to sort by/remove columns respectively.

Preview:
image

- It uses fontawesome (free) 6.7.1 for the icons, which might be questionable but this is the best I know
- My JS skills are sub-zero, so feel free to suggest better handling if I missed something


serveradmin/servershell/static/webfonts/fa-solid-900.ttf
kofrezo, line 1: Please add the `webfonts` directory to the [setup.py](https://github.com/innogames/serveradmin/blob/main/setup.py#L47) so it is included in the package:

serveradmin/servershell/static/js/servershell/result.js
```diff
@@ -15,7 +15,25 @@ servershell.update_result = function() {
...
+                servershell.commands.orderby(attribute);
+            }),
+            $(`<a href="#" class="attr-tooltip" title="Remove ${attribute} column">`).append(
+                $('<i class="fa-solid fa-xmark">')
+            ).click(function(e) {
+                e.preventDefault();
+                servershell.commands.attr(attribute);
+            }),
+            // Add 1 space after minus icon to visually separate the columns
+            $('<span>').html('&nbsp;')
```
kofrezo, line 34: This is preferably done via CSS (margin right).

serveradmin/servershell/static/webfonts/fa-solid-900.ttf
kofrezo: This has to go to `serveradmin.servershell` as you put the web fonts there.

kofrezo (approved):

PR #384, opened 2025-01-21
pdns: update SQL "domains" view to be compatible with pdns-server 4.7

brainexe:
https://doc.powerdns.com/authoritative/upgrading.html#to-4-7-0


kofrezo:
Already handled by #385

PR #388, opened 2025-03-10
Fix is_unique_ip() logic

TuxPowered42:


serveradmin/serverdb/models.py
```diff
@@ -126,6 +127,9 @@ def is_unique_ip(
     # only withing the same attribute id. That means different hosts can have
     # the same IP address as long as it is in different attributes.
 
+    duplicates = []
+    object_id = server.server_id
```
kofrezo, line 131: Why passing the whole server object when we only needing the object_id ?

serveradmin/serverdb/models.py
```diff
@@ -100,8 +100,9 @@ def is_ip_address(ip_interface: Union[IPv4Interface, IPv6Interface]) -> None:
 
 def is_unique_ip(
     ip_interface: Union[IPv4Interface, IPv6Interface],
-    object_id: int,
-    attribute_id: str = None,
+    server: "Server",
```
kofrezo, line 103: Shouldn't this be Server without double quotes ?

serveradmin/serverdb/models.py
TuxPowered42: This commit also improves error message where `server.hostname` is used, which greatly helps when committing multiple objects. Hence changing the attribute from `object_id` to `server`.

serveradmin/serverdb/models.py
TuxPowered42: The class is defined in the same but later. We need a forward reference.

lamaral (commented):
The logic change LGTM.
The formatting I'll leave up to Daniel.

serveradmin/serverdb/models.py
kofrezo: Okay fair enough. 

Just be aware that with that you force the method to be aware of the structure of the Server model implicitly and while it is currently the case we might not always keep hostname as unique value but only object_id.

serveradmin/serverdb/models.py
kofrezo: Ah true.

kofrezo (approved):

PR #392, opened 2025-03-21
Remove intern_ip support from supernet calculation

TuxPowered42:
This makes it possible to calculate AF-unaware supernets over any attribute. Update tests to provide both AF-aware and AF-unaware supernets, name them the same way we use them at InnoGames for simplicity.

### DIFF
--- a/serveradmin/serverdb/query_materializer.py
+++ b/serveradmin/serverdb/query_materializer.py
@@ -8,11 +8,9 @@ Copyright (c) 2019 InnoGames GmbH
 # a good idea to refactor this by using more top level functions instead of
 # object methods.
 
-import itertools
+import logging
 
 from ipaddress import IPv4Address, IPv6Address
-from django.db.models import OuterRef, Subquery, AutoField
-
 from adminapi.dataset import DatasetObject
 from serveradmin.serverdb.models import (
     Servertype,
@@ -21,8 +19,9 @@ from serveradmin.serverdb.models import (
     Server,
     ServerAttribute,
     ServerRelationAttribute,
-    ServerInetAttribute,
 )
+
+logger = logging.getLogger(__package__)
 
 
 class QueryMaterializer:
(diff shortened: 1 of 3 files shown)


lamaral (approved):

serveradmin/serverdb/query_materializer.py
```diff
@@ -197,62 +196,59 @@ def _add_domain_attribute(self, attribute, servers):
...
-
-        supernets = Server.objects.filter(server_id__in=attrs.values("supernet_id"))
-        supernets_by_id = {s.server_id: s for s in supernets}
-
-        for a in attrs:
-            if a.supernet_id in supernets_by_id:
-                self._server_attributes[a.server][attribute] = supernets_by_id[
-                    a.supernet_id
-                ]
+        q = """
```
kofrezo, line 199: Care to share what this query is doing and why we by pass the query set interface in a comment ?

serveradmin/serverdb/query_materializer.py
```diff
@@ -197,62 +196,59 @@ def _add_domain_attribute(self, attribute, servers):
...
+            JOIN attribute AS net_attr ON (net_attr.attribute_id = net_addr.attribute_id)
+            JOIN attribute AS host_attr ON (host_attr.attribute_id = host_addr.attribute_id)
+            WHERE
+                net.servertype_id = %(target_servertype)s
+                AND host.server_id = any(%(hosts)s)
+        """
 
-    def _add_supernet_attribute_intern_ip(self, attribute, servers):
-        servers_in_1, servers_in_2 = itertools.tee(servers)
+        if attribute.inet_address_family:
```
kofrezo, line 215: In which case could this be blank ?

serveradmin/serverdb/query_materializer.py
TuxPowered42: It's not a comment, it's a multiline string.

kofrezo: Yes I mean maybe you can describe the query's intention in a comment not everybody might look at the SQL and find it obvious what it is doing.

kofrezo: And more importantly why.

serveradmin/serverdb/query_materializer.py
TuxPowered42: For example `route_network` attribute has no address family. For this supernet the code should check all inet type attributes and ensure that they point to the same supernet. `route_networks` are dual stack.

On the other hand `provider_network_ipv4` and `provider_network_ipv6` have address family. A host can belong to multiple provider networks, like a local subnet matching on `10.0.0.0/8` for Legacy IP and some a IPv6 subnet, like `2001:db8::/32`.

serveradmin/serverdb/query_materializer.py
```diff
@@ -197,62 +196,59 @@ def _add_domain_attribute(self, attribute, servers):
...
-        for a in attrs:
-            if a.supernet_id in supernets_by_id:
-                self._server_attributes[a.server][attribute] = supernets_by_id[
-                    a.supernet_id
-                ]
+        q = """
+            SELECT
+                net.server_id, net.hostname, net.intern_ip, net.servertype_id,
+                net_attr.inet_address_family, host_attr.inet_address_family,
+                host.server_id AS host_server_id, host_attr.attribute_id AS host_attr_name
```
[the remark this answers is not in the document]
TuxPowered42: Done.

serveradmin/serverdb/query_materializer.py
TuxPowered42: Done.

kofrezo (approved):

PR #395, opened 2025-04-22
Make the IP Address Chooser configurable

TuxPowered42:


serveradmin/servershell/views.py
```diff
@@ -490,15 +494,12 @@ def choose_ip_addr(request):
...
 
-    # TODO: This is specific to our data model, we should get it independent
-    network_query = Query(
-        {'intern_ip': network, 'servertype': Not('provider_network')},
-        ['intern_ip', 'servertype'])
+    network_query = Query({'intern_ip': network}, ['intern_ip'])
 
-    return TemplateResponse(request, 'servershell/choose_ip_addr.html', {
+    return TemplateResponse(request, 'servershell/choose_ip_address.html', {
         'ip_addrs': islice(network_query.get_free_ip_addrs(), 50)})
```
kofrezo, line 503: This must have been broken before. The `get_free_ip_addrs` method is not available in the class `serveradmin.dataset.Query` but only in `adminapi.dataset.Query`. 

If you do not rely on it (I guess, otherwise you would have noticed) you can ignore it.

PR #397, opened 2025-05-14
Return HTTP status code 400 on ApiError

kofrezo:
Fix ApiError exceptions raised result in a HTTP status code 500 and are retried.


kofrezo:
This only fixes the API. The Servershell still uses different endpoint. This is something to clean up to use the same endpoint and error handling.

TuxPowered42 (approved):

PR #398, opened 2025-05-15
serverdb: Add a bit of verbosity

seqizz:
Especially in the Puppet context, it is better to have the failed hostname printed within the error.


serveradmin/serverdb/query_committer.py
```diff
@@ -759,7 +759,7 @@ def _validate_real_attributes(servertype, real_attributes):     # NOQA: C901
 def _insert_server(hostname, intern_ip, servertype, attributes):
 
     if Server.objects.filter(hostname=hostname).exists():
-        raise CommitError('Server with that hostname already exists')
+        raise CommitError(f'Server with hostname "{hostname}" already exists')
```
kofrezo, line 762: Longterm this should be object_id but in this context we have the hostname so this is easiest.

kofrezo (approved):

PR #404, opened 2025-07-22
Fix UnicodeDecorError on HTTP status code 4xx

kofrezo:


adminapi/request.py
```diff
@@ -130,7 +130,7 @@ def calc_app_id(auth_token):
 def send_request(endpoint, get_params=None, post_params=None):
     for retry in reversed(range(Settings.tries)):
         request = _build_request(endpoint, get_params, post_params)
-        response = _try_request(request, retry)
+        response = _try_request(request, bool(retry))
```
TuxPowered42, line 133: It took me a moment to realize what it does. Why not simpler code like `retry != 0`?

kofrezo: Yeah I like it. That is even better to understand.

PR #407, opened 2025-08-08
Servershell: More graceful error handling

kofrezo:


kofrezo:
I did not understood the error value on of the promise catch function in the last PR. This does check for the actual error and handles some more things.

brainexe (approved):

PR #427, opened 2026-02-10
Allow a selection of target servertypes for attributes

kofrezo:
It is currently only possible to select one or none (all) target servertype for attribtues of type relation, supernet and domain.

We have a use case where we want to allow some servertypes as target but not all.

- serverdb.models.Attribute: Replace FK with ManyToManyField
- serverdb.sql_generator: Replace '=' with 'IN' operator
- serverdb.query_materializer: Replace '=' with 'ANY' operator
- serverdb.query_materializer: Use Django '__in' instead of '=' for ORM
- serverdb.admin: Add filter_horizontal widget, remove from readonly_fields
- serverdb.admin: Explain what happens if no servertype is selected
- serverdb.forms: Adapt validation from is-not-None to queryset .exists()
- Add migration to adjust schema and migrate existing data
- Adjust fixtures for tests
- Adjust db/integrity.sql: Updated to query M2M join table (likely orphaned)

### DIFF
--- a/serveradmin/serverdb/sql_generator.py
+++ b/serveradmin/serverdb/sql_generator.py
@@ -25,6 +25,7 @@ from adminapi.filters import (
     StartsWith,
     Not,
 )
+from serveradmin.serverdb import models
 from serveradmin.serverdb.models import (
     Server,
     ServerAttribute,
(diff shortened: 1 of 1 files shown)


serveradmin/serverdb/sql_generator.py
```diff
@@ -207,6 +207,17 @@ def _containment_filter_template(attribute, filt):
     return template.format(_raw_sql_escape(value))
 
 
+def _target_servertype_sql(alias, attribute):
```
brainexe: could we use typehint at least for new code, else it's alwas hard to see is the attribute is the plain string or some django model.

PR #442, opened 2026-05-04
serverdb: Unregister Server model from admin interface

seqizz:
Helped immensely in techerror, might help clicking individual server elements in serveradmin admin interface

### DIFF
--- a/serveradmin/serverdb/admin.py
+++ b/serveradmin/serverdb/admin.py
@@ -108,4 +108,3 @@ class AttributeAdmin(admin.ModelAdmin):
 
 admin.site.register(Servertype, ServertypeAdmin)
 admin.site.register(Attribute, AttributeAdmin)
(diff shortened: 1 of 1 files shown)


serveradmin/serverdb/admin.py
```diff
@@ -49,6 +49,7 @@ def get_readonly_fields(self, request, obj=None):
 class ServerRelationAttributeInline(admin.TabularInline):
     model = ServerRelationAttribute
     fk_name = 'server'
+    raw_id_fields = ('value',)
```
kofrezo: Lets not do this. This is technically valid and a good way to speed up loading foreign key contains but we rely on our custom API (query_committer) to enforce validation, authorization etc. so people should really not edit servers in the admin panel directly but use the Servershell.

Lets make this clear by removing the line to register the Server model below.

kofrezo (approved):

PR #445, opened 2026-06-03
serverdb: Fix Any() filter on boolean attributes

kofrezo:
Booleans have no value column; presence of a row means true. Without
this, Any() built an IN (...) comparison against a non-existent column,
making queries like Any(True, False) fail instead of matching servers.


kofrezo:
Fixes errors when using Any on boolean attributes:

```
ProgrammingError
column sub.value does not exist
LINE 1: ...ver_id AND sub.attribute_id = 'no_monitoring' AND sub.value ...
````

abeglu (approved):

PR #448, opened 2026-06-11
Remove obsolete dependency pexpect

kofrezo:
This was probably added for the internal Django apps now living in an extra git repository.


brainexe (approved):

kofrezo:
Force updated to regenerate Pipefile.lock.

PR #453, opened 2026-07-01
Improve security of DB dump script

kofrezo:
- Do not prompt for password anymore
- Disable production accounts after dump
- Change tokens after dump to prevent accidental leak

### DIFF
--- a/dump.sh
+++ b/dump.sh
@@ -1,8 +1,20 @@
 #!/bin/bash -e
 
-psql -h localhost -U "$POSTGRES_USER" -d postgres -c "drop database if exists $POSTGRES_DB;"
-psql -h localhost -U "$POSTGRES_USER" -d postgres -c "create database $POSTGRES_DB;"
+export PGPASSWORD="$POSTGRES_PASSWORD"
+
+sql() {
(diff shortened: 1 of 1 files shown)


dump.sh
```diff
@@ -1,8 +1,20 @@
...
     | gunzip -c \
-    | psql -h localhost -U "$POSTGRES_USER" -d "$POSTGRES_DB"
\ No newline at end of file
+    | psql -h localhost -U "$POSTGRES_USER" -d "$POSTGRES_DB"
+
+
+# To have less of a risk of accessing production and preventing token leaks
+# we better disable applications and change tokens
+sql "$POSTGRES_DB" 'update apps_application set disabled = true;'
+sql "$POSTGRES_DB" "update apps_application set auth_token = substr(md5(random()::text || id::text), 1, 25);"
```
TuxPowered42: My personal dump excludes the application table:
    --exclude-table-data='serverdb_*' \
    --exclude-table-data='apps_application' \

Do we need it imported at all?

kofrezo: True! I did not though of this despite we are doing it already above 🤦. 

Yes this is better then we don't download the data at all in the first place!

TuxPowered42 (approved):

PR #454, opened 2026-07-01
serverdb: Fix false commit conflict on datetimes

kofrezo:
_validate_commit compared the current value via json_encode_extra() 
against the old value via str(). For datetime attributes these never 
match, since str() keeps microseconds and the "+00:00" tz form while
json_encode_extra() drops both.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -68,7 +68,7 @@ jobs:
         run: |
           # This will use Python's standard unit test discovery feature.
           pipenv run python -m unittest discover adminapi -v
-          pipenv run python -Wall -m serveradmin test --noinput --parallel=1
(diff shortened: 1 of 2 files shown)


kofrezo:
This should fix bugs such as:
```
serveradmin.serverdb.query_committer.CommitNewerData: ["Newer data available for attribute [(9, 'last_updated', datetime.datetime(2026, 7, 1, 13, 25, 31, 987683, tzinfo=datetime.timezone.utc))]"]
```

seqizz (approved):
As far as I understand this will solve most (maybe all) errors we're getting for backup backend at least, since we pass datetime object, now being normalized. If I still keep seeing issues, I can dig a bit more, maybe I am doing something extra wrong (while passing both backup_state and backup_datetime attrs in same commit).

PR #456, opened 2026-07-06
Add popstate to recall history on state change

YannikSc:
- This will cause the "Search" field to update the term + a trigger a search when the browser history is changed either back or forward


kofrezo (approved):

kofrezo:
This allows to use the browser back (and forward) button to jump back to older (or forward to newer) searches.

PR #457, opened 2026-07-09
servershell: speed up result table rendering

seqizz:
Selecting and working on many objects is slow because of the current per-row jQuery construction and per-cell linear scans. This commit tries to speed up via multiple changes. Feels much snappier but don't have a solid benchmark yet.

Mostly work of Opus, still sending all the inputs through escape_html to avoid XSS.

- Rewrite update_result/get_row_html to build the body as one HTML string (single innerHTML) instead of per-element jQuery construction.
- Precompute O(1) lookups per redraw (object/attribute/deleted/changed) instead of linear scans per cell.
- Snapshot the servershell Proxy properties used in the loop (each access dispatches events).
- Event delegation for inline-edit and relation links instead of per-cell/per-link handlers.
- set_selected O(rows) instead of O(selected * rows).
- Fix delete to redraw once instead of once per selected object.

### DIFF
--- a/serveradmin/servershell/static/js/servershell/command.js
+++ b/serveradmin/servershell/static/js/servershell/command.js
@@ -445,9 +445,11 @@ servershell.commands = {
             // Avoid duplicate deletion ...
             if (!servershell.to_commit.deleted.includes(object_id)) {
                 servershell.to_commit.deleted.push(object_id);
-                servershell.update_result();
             }
         });
+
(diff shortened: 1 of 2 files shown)


kofrezo (approved):
Did some local tests. Looks good to me. I cannot understand every single change but I get the concept of instead of triggering a change for every action and working on a copy and applying it afterwards and that looks sane.

We still need to rebase though (tested with a rebased version)

PR #461, opened 2026-07-31
Fix version still displayed based on adminapi

kofrezo:
Serveradmin and Adminapi both now have independent versions. When seperating the packages this was missed to be adjusted.


seqizz (approved):
Since serveradmin depends on adminapi (afaik) does it make sense to show both versions? Or their versions are always coupled?
LGTM though, makes more sense to show serveradmin.

PR #464, opened 2026-09-04
Rule based logging of Queries based on conditions

kofrezo:
This change allows to define rules with that contain one or more conditions to log some queries.

If, for example, one wants to log all queries of an application that query attribute X one can do so now.


kofrezo:
This is mostly AI generated with Claude but I reviewed all code and did a few adjustments manually. It looks good enough to me for a app that does "just" debugging.

brainexe (approved):

PR #467, opened 2026-09-18
inspect: order multi attributes

brainexe:
having views with many unsorted multi-attributes are a bit hard to search
bad example:
image


kofrezo (changes requested):
It should suffice to sort them in the [views.py](https://github.com/innogames/serveradmin/blob/main/packages/serveradmin/serveradmin/servershell/views.py#L366) so we don't have to add another template tag.

kofrezo (approved):

PR #470, opened 2026-09-22
Improve performance of query materializer

kofrezo:
This PR contains  changes reducing the runtime of queries returning thousands of objects.

- Defer intern_ip if not requested - intern_ip is a "special" attribute that was always present in the past
- Fetch supernets once instead of per row
- Map server attributes by server_id (plain int) instead of Server objects (hash function)
- Hash Attribute by attribute_id directly 
- Read stored attribute rows as tuples


kofrezo:
This code has been written with the assistance of LLMs but manually reviewed and tested. I find the changes meaningful and understandable. Two commits are questionable in thus as they make changes to the standard way Django works and can be dropped if they cause any problems.

brainexe (approved):
manual tests passed, code looks valid ✔️

PR #471, opened 2026-09-22
Improve SQL generator by grouping "exists" statements better

kofrezo:
This PR contains improvements to the sql_generator module to better group exists for related attributes allowing the
Postgres query planner to filter rows more efficient.


kofrezo:
This code has been written with assistance of an LLM. I tested and verified and reviewed them myself. The improvements especially for bigger queries are quite impressive I saw queries dropping from 40 to 5 seconds runtime.

brainexe (approved):
code looks valid + did some local tests and it still produced the expected results

